PR #7, opened 2025-11-03
Integrations plus

IEvangelist:
- Remove "What's new in Aspire 9.5" and replace it with "What's new in Aspire 13".
- Add placeholder bits for integrations with links to Learn.
- Add stubs for porting integrations.
- Port SQL article.

More integration cards end up with a docs icon:

image

More complete TOC, with stub articles and links back to Learn:

image

Selecting the link in the placeholder takes the reader to the Learn article.

### DIFF
--- a/src/frontend/src/content/docs/whats-new/aspire-9-5.mdx
+++ b/src/frontend/src/content/docs/whats-new/aspire-9-5.mdx
@@ -1,0 +1,1705 @@
+---
+title: What's new in Aspire 9.5
+description: Learn what's new in Aspire 9.5.
+sidebar:
+  label: Aspire 9.5
+  order: 10
+tableOfContents:
+  minHeadingLevel: 2
+  maxHeadingLevel: 2
+---
+
+📢 Aspire 9.5 is the next minor version release of Aspire. It supports:
+
+- .NET 8.0 Long Term Support (LTS)
+- .NET 9.0 Standard Term Support (STS)
+- .NET 10.0 Preview 6
+
+import { Icon } from '@astrojs/starlight/components';
+
+If you have feedback, questions, or want to contribute to Aspire, collaborate with us on [<Icon name="github" class='d-inline' /> GitHub](https://github.com/dotnet/aspire) or join us on our new [<Icon name="discord" class='d-inline' /> Discord](https://aka.ms/aspire-discord) to chat with the team and other community members.
+
+It's important to note that Aspire releases out-of-band from .NET releases. While major versions of Aspire align with major .NET versions, minor versions are released more frequently. For more information on .NET and Aspire version support, see:
+
+- [.NET support policy](https://dotnet.microsoft.com/platform/support/policy): Definitions for LTS and STS.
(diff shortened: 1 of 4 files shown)


src/frontend/src/content/docs/integrations/cloud/azure/overview.mdx
```diff
@@ -0,0 +1,12 @@
+---
+title: Azure integrations overview
+---
+
+import Placeholder from "@components/Placeholder.astro";
```
captainsafia, line 5: Is this placeholder going to link out to the Azure-specific resources like Azure Storage and CosmosDB?

IEvangelist: Eventually, but for now, we'll keep it simple. I didn't want to create all the stubs for every Azure integration...

src/frontend/redirects.mjs
```diff
@@ -0,0 +1,9 @@
+export const redirects = {
+    // Install scripts
+    '/install.sh': 'https://aka.ms/aspire/get/install.sh',
```
joperezr: NIT: Should we remove these install.sh/ps1 redirects since we aren't really using them anymore?

src/frontend/redirects.mjs
```diff
@@ -0,0 +1,9 @@
...
+    '/install.ps1': 'https://aka.ms/aspire/get/install.ps1',
+    
+    // Docs redirects
+    '/integrations/postgres': '/integrations/databases/postgres/',
```
joperezr, line 3: Can you help me understand why we need these two?

src/frontend/src/content/docs/whats-new/aspire-9-5.mdx
```diff
@@ -1,1702 +0,0 @@
----
-title: What's new in Aspire 9.5
```
joperezr, line 2: Not a strong preference, but should we keep these around in case people want to see what was added in previous versions?

src/frontend/redirects.mjs
IEvangelist: Yes, today - aspire.dev has two published integration docs:

image

- [Postgres](https://aspire.dev/integrations/postgres/) 
- [RabbitMQ](https://aspire.dev/integrations/rabbitmq/)

When we were first shipping these, we hadn't decided whether or not to nest the content into categories. We've since decided to do so, and as such we need these redirects as these files have moved.

src/frontend/src/content/docs/whats-new/aspire-9-5.mdx
IEvangelist: It felt like noise, since it's already published on Microsoft Learn. If we want to, we can port over all of those as part of the migration.

src/frontend/redirects.mjs
IEvangelist: I can remove them.

src/frontend/src/content/docs/whats-new/aspire-9-5.mdx
joperezr: yeah that might be a good idea. no need to block this PR on that though.

PR #41, opened 2025-11-13
Fix shuffle bug in language cards and update excluded packages

IEvangelist:
Thanks to @edandersen for [finding this](https://youtu.be/zLgHnFQLXUQ?t=260). In this PR:

- Fix shuffle functionality, that was only happening during server-side (build time). Made shuffle client-side, which was the original intention.
- Add .NET MAUI icon and entry for integrations.
- Add a few more packages to exclude from gallery.


edandersen:
Blimey, I thought it must have been a bug after all!

PR #120, opened 2025-12-05
Add cookie consent for compliance

IEvangelist:
Our analytics engine `1DS` is now aligned with cookie compliance, and will not run until the user opts-in (consents) to cookies:

The site, once updated will display the pop-up in the bottom-right corner:

image

After they either accept or reject in will no longer be displayed. Then, we added a cookie icon to the top nav, that toggles the cookie preferences sidebar:

image

In addition, this PR also refactors how some config-based files are organized and adds a safety net to ensure data files are available.


src/frontend/config/head.attrs.ts
```diff
@@ -0,0 +1,64 @@
...
+    { tag: 'link', attrs: { rel: 'icon', type: 'image/png', href: '/favicon-96x96.png', sizes: '96x96' } },
+    { tag: 'link', attrs: { rel: 'shortcut icon', href: '/favicon.ico' } },
+    { tag: 'link', attrs: { rel: 'apple-touch-icon', sizes: '180x180', href: '/apple-touch-icon.png' } },
+    { tag: 'meta', attrs: { name: 'apple-mobile-web-app-title', content: 'Aspire' } },
+    { tag: 'link', attrs: { rel: 'alternate', type: 'application/rss+xml', title: 'Aspire Docs RSS', href: '/rss.xml' } },
+
+    // Analytics scripts
+    {
+        tag: 'script', attrs: {
+            type: 'text/plain',
```
captainsafia, line 50: Should this be `text/javascript`?

IEvangelist: No, this is how the cookie consent works. The script tag is plain text to start, until the user consents, then it's JS.

PR #121, opened 2025-12-05
Set immutable cache headers for _astro/ files at build time

javiercn:
This PR uses the `UpdateStaticWebAssetEndpoints` MSBuild task to set immutable cache headers for Astro's fingerprinted assets (`_astro/` files) at build time.

## Changes
- Added `Caching.targets` with a target that runs after `ResolveProjectStaticWebAssets`
- Uses the `Replace` operation to change `Cache-Control: no-cache` to `max-age=31536000, public, immutable` for `_astro/` endpoints
- Updated `StaticHost.csproj` to import the new targets file

## Why
Astro generates fingerprinted assets in the `_astro/` folder. These files have content hashes in their names, making them safe to cache indefinitely with immutable headers.

### DIFF
--- a/src/statichost/StaticHost/Caching.targets
+++ b/src/statichost/StaticHost/Caching.targets
@@ -16,7 +16,6 @@
       <_CacheHeaderOperation Include="Replace">
         <UpdateTarget>Header</UpdateTarget>
         <Name>Cache-Control</Name>
-        <Value>no-cache</Value>
         <NewValue>max-age=31536000, public, immutable</NewValue>
       </_CacheHeaderOperation>
     </ItemGroup>


src/statichost/StaticHost/Caching.targets
```diff
@@ -0,0 +1,34 @@
...
+    <UpdateStaticWebAssetEndpoints
+      EndpointsToUpdate="@(_AstroEndpoints)"
+      AllEndpoints="@(StaticWebAssetEndpoint)"
+      Operations="@(_CacheHeaderOperation)">
+      <Output TaskParameter="UpdatedEndpoints" ItemName="_UpdatedAstroEndpoints" />
+    </UpdateStaticWebAssetEndpoints>
+
+    <!-- Update the endpoint list -->
+    <ItemGroup>
+      <StaticWebAssetEndpoint Remove="@(_UpdatedAstroEndpoints)" />
```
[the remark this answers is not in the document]
javiercn: Doesn't matter. They both have the same path and using _UpdatedAstroEndpoints is the correct thing.

src/statichost/StaticHost/Caching.targets
```diff
@@ -0,0 +1,34 @@
+<Project>
+  <!-- Set immutable cache headers for _astro/ files at build time -->
+  <Target Name="SetAstroImmutableCacheHeaders" AfterTargets="ResolveProjectStaticWebAssets">
```
javiercn:
```suggestion
  <Target Name="SetAstroImmutableCacheHeaders" AfterTargets="ResolvePublishStaticWebAssets">
```

src/statichost/StaticHost/Caching.targets
```diff
@@ -0,0 +1,34 @@
+<Project>
+  <!-- Set immutable cache headers for _astro/ files at build time -->
+  <Target Name="SetAstroImmutableCacheHeaders" AfterTargets="ResolvePublishStaticWebAssets">
```
eerhardt:
```suggestion
  <Target Name="SetAstroImmutableCacheHeaders" AfterTargets="ResolvePublishCompressedStaticWebAssets">
```

src/statichost/StaticHost/Caching.targets
```diff
@@ -0,0 +1,34 @@
+<Project>
+  <!-- Set immutable cache headers for _astro/ files at build time -->
+  <Target Name="SetAstroImmutableCacheHeaders" AfterTargets="ResolvePublishCompressedStaticWebAssets">
```
eerhardt:
```suggestion
  <!--
  Set immutable cache headers for _astro/ files at build time.
  Anything served from the /_astro/ directory can be cached forever
  https://starlight.astro.build/environmental-impact/#caching
  -->
  <Target Name="SetAstroImmutableCacheHeaders" AfterTargets="ResolvePublishCompressedStaticWebAssets">
```

src/statichost/StaticHost/Caching.targets
```diff
@@ -0,0 +1,38 @@
...
+      <_AstroEndpoints Include="@(StaticWebAssetEndpoint)" 
+                       Condition="$([System.String]::Copy('%(Identity)').StartsWith('_astro/'))" />
+    </ItemGroup>
+
+    <!-- Define the operation: Replace Cache-Control header with immutable value -->
+    <ItemGroup>
+      <_CacheHeaderOperation Include="Replace">
+        <UpdateTarget>Header</UpdateTarget>
+        <Name>Cache-Control</Name>
+        <Value>no-cache</Value>
```
eerhardt: One issue is that some of the assets have a Cache-Control that isn't `no-cache`. For example, .svg files. This won't update those, but instead some assets get `max-age=31536000` and some get `max-age=3600, must-revalidate`

javiercn: We can remove the `value` and it will match only on name I believe

eerhardt:
```suggestion
```

This seems to fix it for me locally.

PR #138, opened 2025-12-10
Authoring custom integrations: hosting, client & securing communications between the two

IEvangelist:


.github/prompts/migrate-from-learn.prompt.md
```diff
@@ -216,7 +216,7 @@ Copy-Item "source" "destination"
 5. What components need translation?
 6. What messaging improvements are desired?
 7. Are there dependencies on other articles?
-8. Does similar content already exist in aspire.dev?
+8. Does similar content already exist in aspire.dev
```
eerhardt: Why the removal of the `?`?

IEvangelist: That's a good question. I think AI might have updated that...I know I didn't explicitly remove it.

PR #162, opened 2025-12-14
Update certificate trust article to cover HTTPS endpoint configuration as well

danegsta:
This adds documentation and examples of the new HttpsCertificate API for configuring resource HTTPS/TLS endpoints that's added in 13.1 in addition to the certificate trust APIs from 13.0.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -441,3 +441,6 @@ FodyWeavers.xsd
 *.msix
 *.msm
 *.msp
+
+# Mac folder metadata
+.DS_Store
--- a/src/frontend/frontend.esproj
+++ b/src/frontend/frontend.esproj
@@ -4,5 +4,6 @@
     <ShouldRunNpmInstall>false</ShouldRunNpmInstall>
     <!-- Disables running the build (or compile) script located on package.json during Build -->
     <ShouldRunBuildScript>false</ShouldRunBuildScript>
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -441,3 +441,6 @@ FodyWeavers.xsd
 *.msix
 *.msm
 *.msp
+
+# Mac folder metadata
+.DS_Store
```
danegsta, line 446: This suppresses a metadata file that gets created whenever a folder is opened in the file explorer on MacOS. Not directly related to this PR, but I had one accidentally make it into the initial commit and wanted to prevent others from hitting the same.

src/frontend/frontend.esproj
```diff
@@ -4,5 +4,6 @@
     <ShouldRunNpmInstall>false</ShouldRunNpmInstall>
     <!-- Disables running the build (or compile) script located on package.json during Build -->
     <ShouldRunBuildScript>false</ShouldRunBuildScript>
+    <IntermediateOutputPath>$(MSBuildProjectDirectory)/obj/$(Configuration)</IntermediateOutputPath>
```
danegsta, line 7: Default IntermediateOutputPath for esproj seems to be Windows centric; it's got a `\` in the path which doesn't play nice with Mac or Linux and breaks the .gitignore for obj. This ensures paths will behave properly on all OS.

IEvangelist: This might be a good XML comment to leave in this file for later, no?

PR #210, opened 2026-01-06
Reference bits

IEvangelist:
Refactor our reference topic to include API reference. Today, we'll just expose an API search that will link out to `learn.microsoft.com`, but eventually we will host API docs on `aspire.dev`. This is an initial step toward that long-term goal.

We currently support a version dropdown, from the learn API, and we also rely on the learn API for searching APIs. Again, this is just a first step. The initial part of the video didn't include the descriptions that were coming back from the APIs, the later part does. We support search, versioned search, and loading more results...

https://github.com/user-attachments/assets/5a63539e-190f-4ec3-a664-e6b343f8b850

### DIFF
--- a/src/frontend/src/components/AspireApiReferenceSearch.astro
+++ b/src/frontend/src/components/AspireApiReferenceSearch.astro
@@ -164,6 +164,17 @@ try {
     }
 
     class ApiSearchController {
+        // Constants
+        private readonly BASE_API_URL = 'https://learn.microsoft.com';
+        private readonly BASE_API_ORIGIN = 'https://learn.microsoft.com';
+        private readonly LEARN_API_VERSION = '0.2';
+        private readonly DEFAULT_LOCALE = 'en-us';
+        private readonly PAGE_SIZE = 25;
+        private readonly DEBOUNCE_DELAY_MS = 250;
+        private readonly MINIMUM_LOADING_TIME_MS = 500;
+        private readonly DISPLAY_NAME_LONG_THRESHOLD = 80;
+        private readonly DEFAULT_MONIKER = 'dotnet-aspire-13.0';
+
         private container: HTMLElement;
         private input: HTMLInputElement;
         private resultsContainer: HTMLElement;
@@ -175,7 +186,7 @@ try {
         private debounceTimer: number | null = null;
         private abortController: AbortController | null = null;
         private loadingStartTime: number = 0;
-        private apiVersion = 'dotnet-aspire-13.0';
+        private apiVersion = this.DEFAULT_MONIKER;
         private hasMoreResults: boolean = false;
         private currentSearchTerm: string = '';
         private currentResults: ApiSearchResult[] = [];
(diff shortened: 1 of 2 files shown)


IEvangelist:
Hey @copilot, please navigate to this new browser page:

- Reject all cookies
- Take a screenshot
- Then enter a search term of "AzureOpenAISettings" and take another screenshot

I really want to see what this looks like, and please do this in dark theme.

src/frontend/src/content/docs/reference/api/browser.mdx
```diff
@@ -0,0 +1,18 @@
...
+editUrl: false
+tableOfContents: false
+---
+
+import AspireApiReferenceSearch from '@components/AspireApiReferenceSearch.astro';
+
+Welcome to the Aspire API browser — your one-stop shop for all `public` APIs from Aspire. To get started, simply enter your search terms in the search box below. You can search by API names, types, members, or keywords to quickly find the information you need.
+
+:::note
+We're actively exploring ways to improve the API browsing experience. Today, we still host APIs on [learn.microsoft.com](https://learn.microsoft.com/dotnet/api/?view=dotnet-aspire-13.0), but we're working towards integrating more features directly into this browser. Stay tuned for updates!
```
eerhardt, line 15:
```suggestion
We're actively exploring ways to improve the API browsing experience. Today, we still host APIs on [learn.microsoft.com](https://learn.microsoft.com/dotnet/api/?view=dotnet-aspire-13.1), but we're working towards integrating more features directly into this browser. Stay tuned for updates!
```

?

src/frontend/src/components/AspireApiReferenceSearch.astro
```diff
@@ -0,0 +1,1178 @@
...
+                    return;
+                }
+                await this.ensureMinimumLoadingTime();
+                this.renderError(error instanceof Error ? error.message : "An unknown error occurred");
+            } finally {
+                this.setLoading(false);
+            }
+        }
+
+        private async ensureMinimumLoadingTime(): Promise<void> {
```
eerhardt, line 365: Why is this necessary?

eerhardt (approved):

src/frontend/src/content/docs/reference/api/browser.mdx
IEvangelist: That's not the moniker. It's 13.0 for the latest 13 bits, when 13.2.0 is shipped this will still be 13.0. Don't ask. 😝

src/frontend/src/components/AspireApiReferenceSearch.astro
IEvangelist: To avoid a jitter in the UX, I impose an 500ms load time. This makes the UX more seamless.

eerhardt: Would it make sense adding a comment for this?

PR #240, opened 2026-01-12
Add GitHub stats data for Aspire projects

IEvangelist:
- Introduced a new JSON file containing statistics for various Aspire-related repositories.
- Included details such as repository name, star count, description, license information, and repository URL.


src/frontend/src/data/aspire-integration-names.json
```diff
@@ -0,0 +1,141 @@
+[
+  "Aspire.Azure.AI.Inference",
```
eerhardt: Why do we need this separate document? Isn't this duplicate information with what's in aspire-integrations.json?

.github/workflows/dependency-updates.yml
```diff
@@ -44,27 +44,18 @@ jobs:
...
-            echo "changed=false" >> $GITHUB_OUTPUT
-            echo "No changes detected in integration data" >> $GITHUB_STEP_SUMMARY
-          fi
+      - name: Get current date
+        id: date
+        run: echo "date=$(date +'%d/%m/%y' | sed 's/0\([1-9]\)/\1/g')" >> $GITHUB_OUTPUT
 
       - name: Create Pull Request
-        if: steps.verify-changed-files.outputs.changed == 'true'
         uses: peter-evans/create-pull-request@v6
```
eerhardt: We should probably be using a more "trusted" Create Pull Request task.

[https://github.com/dotnet/aspire/blob/2c9b42360a1e5a0a91f65d64b286c41976537ced/.github/workflows/refresh-manifests.yml#L30](https://github.com/dotnet/aspire/blob/2c9b42360a1e5a0a91f65d64b286c41976537ced/.github/workflows/refresh-manifests.yml#L30)

eerhardt (approved):
:shipit:

src/frontend/src/data/aspire-integration-names.json
IEvangelist: Good catch. I deleted this. My original thinking was using it within update integrations prompt, so that it could be more efficient, rather than loading and parsing the much larger JSON file with all the details. I suppose though, it's just noise.

PR #256, opened 2026-01-15
refactor: reorganize and enhance the "Why HTTPS matters" section in c…

IEvangelist:
…ertificate configuration documentation

Today, this article has two "Overview" headings, we there's an implicit one for the top of the article. I slightly reorganized this to have the actual overview be first, then why HTTPS matters as a subsection.


danegsta (approved):
Change makes sense to me.

PR #261, opened 2026-01-16
feat: implement OneDSTelemetryService for download tracking and add t…

IEvangelist:
…elemetry configuration

### DIFF
--- a/src/statichost/StaticHost/Extensions.cs
+++ b/src/statichost/StaticHost/Extensions.cs
@@ -6,16 +6,6 @@ public static class Extensions
     {
         builder.ConfigureOpenTelemetry();
 
-        // Register 1DS analytics for downloads...
-        builder.Services.AddOpenTelemetry().UseAzureMonitor(static options =>
-        {
-            options.ConnectionString = TelemetryConstants.AzureMonitorConnectionString;
-        });
-        builder.Services.ConfigureOpenTelemetryTracerProvider(static (_, builder) =>
-        {
-            builder.AddSource(TelemetryConstants.AspireDotDevCliDownload);
-        });
-
         builder.Services.AddSingleton<OneDSTelemetryService>();
 
         return builder;
@@ -30,6 +20,12 @@ public static class Extensions
         });
 
         builder.Services.AddOpenTelemetry()
+            .UseAzureMonitor(static options =>
+            {
+                options.ConnectionString = TelemetryConstants.AzureMonitorConnectionString;
+            });
+
+        builder.Services.AddOpenTelemetry()
             .WithMetrics(metrics =>
             {
                 metrics.AddAspNetCoreInstrumentation()
@@ -38,7 +34,10 @@ public static class Extensions
             })
             .WithTracing(tracing =>
             {
-                tracing.AddSource(builder.Environment.ApplicationName)
+                tracing.AddSource(
+                        builder.Environment.ApplicationName,
+                        TelemetryConstants.AspireDotDevSource
+                    )
                     .AddAspNetCoreInstrumentation()
                     .AddHttpClientInstrumentation();
             });
--- a/src/statichost/StaticHost/Telemetry/TelemetryConstants.cs
+++ b/src/statichost/StaticHost/Telemetry/TelemetryConstants.cs
@@ -6,9 +6,9 @@ namespace StaticHost.Telemetry;
 internal static class TelemetryConstants
 {
     /// <summary>
-    /// The activity source name for Aspire Dev CLI download events.
+    /// The activity source name for <c>aspire.dev</c> events.
     /// </summary>
-    public const string AspireDotDevCliDownload = "Aspire.Dev.Cli.Download";
+    public const string AspireDotDevSource = "Aspire.Dev";
 
     /// <summary>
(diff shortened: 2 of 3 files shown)


src/statichost/StaticHost/Extensions.cs
```diff
@@ -10,6 +6,18 @@ public static TBuilder AddServiceDefaults<TBuilder>(this TBuilder builder) where
     {
         builder.ConfigureOpenTelemetry();
 
+        // Register 1DS analytics for downloads...
+        builder.Services.AddOpenTelemetry().UseAzureMonitor(static options =>
+        {
+            options.ConnectionString = TelemetryConstants.AzureMonitorConnectionString;
```
eerhardt: (nit) these first 2 telemetry calls could all go into the existing ConfigureOpenTelemetry and AddOpenTelemetryExporters methods. That's how we intend people to customize this. Ex:

[https://github.com/dotnet/aspire/blob/main/src/Aspire.ProjectTemplates/templates/aspire-servicedefaults/Extensions.cs#L90-L95](https://github.com/dotnet/aspire/blob/main/src/Aspire.ProjectTemplates/templates/aspire-servicedefaults/Extensions.cs#L90-L95)

And the `AddSource` call could go in the existing:

```C#
.WithTracing(tracing =>
            {
                tracing.AddSource(builder.Environment.ApplicationName)
                    .AddAspNetCoreInstrumentation()
                    .AddHttpClientInstrumentation();
            });
```

block.

src/statichost/StaticHost/Telemetry/TelemetryConstants.cs
```diff
@@ -0,0 +1,78 @@
...
+
+    /// <summary>
+    /// Tag names for telemetry data.
+    /// </summary>
+    internal static class Tags
+    {
+        /// <summary>
+        /// Tag indicating the behavior associated with the event.
+        /// </summary>
+        public const string Behavior = "aspire.dev.behavior";
```
eerhardt: Do these tags all need to be prefixed with `aspire.dev`? Isn't it redundant that they are under an `Aspire.Dev.Cli.Download` activity?

src/statichost/StaticHost/Telemetry/TelemetryConstants.cs
```diff
@@ -0,0 +1,78 @@
...
+    /// <summary>
+    /// The Azure Monitor connection string used for 1DS/Application Insights telemetry.
+    /// </summary>
+    public const string AzureMonitorConnectionString = $"InstrumentationKey={InstrumentationKey}";
+
+    /// <summary>
+    /// The instrumentation key used for 1DS/Application Insights telemetry.
+    /// </summary>
+    /// <remarks>
+    /// This key is the same as the one used in the client-side 1ds.js.
```
eerhardt, line 22: Are you sure we don't need a unique one for the server?

src/statichost/StaticHost/Telemetry/TelemetryConstants.cs
```diff
@@ -0,0 +1,78 @@
...
+
+/// <summary>
+/// Contains constants for telemetry tag names and event names used by the <c>aspire.dev</c> static host.
+/// </summary>
+internal static class TelemetryConstants
+{
+    /// <summary>
+    /// The activity source name for Aspire Dev CLI download events.
+    /// </summary>
+    public const string AspireDotDevCliDownload = "Aspire.Dev.Cli.Download";
```
eerhardt: Should we make this Activity Source name more generic in case we need to start tracking more activities in the future? Putting "Download" in it makes it speicfic to one action.

Also - why `Cli`?

src/statichost/StaticHost/Telemetry/TelemetryConstants.cs
IEvangelist: We need it to go to the same destination for aspire.dev telemetry, to my knowledge, no.

src/statichost/StaticHost/Telemetry/TelemetryConstants.cs
IEvangelist: Initially, I was thinking that it was a one-to-one relationship with the activity source name and the activity. I'm super familiar with activity source naming conventions, but I agree - and have adjusted this.

src/statichost/StaticHost/Telemetry/TelemetryConstants.cs
IEvangelist: Fixed, yeah that was just noise:
image

src/statichost/StaticHost/Extensions.cs
```diff
@@ -21,6 +19,12 @@ private static TBuilder ConfigureOpenTelemetry<TBuilder>(this TBuilder builder)
             logging.IncludeScopes = true;
         });
 
+        builder.Services.AddOpenTelemetry()
+            .UseAzureMonitor(static options =>
+            {
+                options.ConnectionString = TelemetryConstants.AzureMonitorConnectionString;
+            });
+
         builder.Services.AddOpenTelemetry()
```
eerhardt, line 28: (nit) can this be chained together in one call to AddOpenTelemetry()?

PR #264, opened 2026-01-20
Footer ux

IEvangelist:
Redesign the footer, added links, use three column layout, improve responsiveness, etc.

### DIFF
--- a/src/frontend/src/components/FooterResources.astro
+++ b/src/frontend/src/components/FooterResources.astro
@@ -14,6 +14,16 @@
         {Astro.locals.t('footer.support')}
       </a>
     </li>
+    <li>
+      <a href="https://github.com/dotnet/aspire/releases" target="_blank" rel="noopener noreferrer">
+        {Astro.locals.t('footer.releases' as any) || 'Releases'}
+      </a>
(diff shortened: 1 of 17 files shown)


src/frontend/src/components/FooterLinks.astro
```diff
@@ -9,62 +9,81 @@ const is404 = Astro.locals.starlightRoute?.entry?.slug === "404";
...
-        padding-top: 2rem;
-		font-size: var(--sl-text-sm);
-	}
+    .footer-column {
+        display: flex;
+        flex-direction: column;
+        gap: 0.5rem;
+    }
+
+    .footer-heading {
```
adamint, line 60: Going to approve but @IEvangelist is the CSS duplication intended?

adamint (approved):
pulled down, looks good. see comment

src/frontend/src/components/FooterLinks.astro
IEvangelist: Yes, in Astro CSS is scoped to a component. We could make it global, but then it gets more messy. This keeps it isolated here, which is easier to maintain.

PR #283, opened 2026-01-22
Update commit link in Footer to point to the correct repository

IEvangelist:


src/frontend/src/components/starlight/Footer.astro
```diff
@@ -35,7 +35,7 @@ const showMap = isHomepage(Astro);
     <div class="footer-bottom">
       <a
         class="commit-link"
-        href={`https://github.com/dotnet/aspire/commit/${import.meta.env.PUBLIC_GIT_COMMIT_ID || ''}`}
+        href={`https://github.com/microsoft/aspire.dev/commit/${import.meta.env.PUBLIC_GIT_COMMIT_ID || ''}`}
```
eerhardt: Why aren't we using `PUBLIC_REPO_URL` here?

https://github.com/microsoft/aspire.dev/blob/9ab891ec5cc8b82d43f3e9590b90af20f6bc8451/src/frontend/scripts/write-git-env.mjs#L42-L45

IEvangelist: Good catch! Fixed...

PR #384, opened 2026-02-07
Document testing with TUnit test runner

Odonno:


src/frontend/src/content/docs/testing/write-your-first-test.mdx
```diff
@@ -50,6 +51,13 @@ dotnet new aspire-mstest -o MSTest.Tests
 dotnet new aspire-nunit -o NUnit.Tests
 ```
 
+</Pivot>
+<Pivot id="tunit">
+
+```bash
+dotnet new aspire-tunit -o TUnit.Tests
```
IEvangelist, line 58: Where does this template come from? Do we need to have the user install one?

IEvangelist (approved):
This PR looks good so far!

src/frontend/src/content/docs/testing/write-your-first-test.mdx
Odonno: It does not exist. I took inspiration from the other test framework. Not sure what to do here.

IEvangelist: Oh, okay. To add this sort of content to the docs — we really need product truth. Ideally, with something like this, we'd start by filing an issue on the product side, proposing that a testing template for TUnit is added. Then from there, assuming it's something that team agrees should be added (and after it's designed, reviewed, approved, and merged) we'd add this content to correspond to the release of said feature.

If it's a community template, that's ok too... but then we'd need install instructions (and probably a discussion about that as well). Does that make sense? I think we'll close this for now.

Odonno: Sure.

PR #601, opened 2026-03-23
Improve API docs build performance and reduce build invalidation

IEvangelist:
## Summary

This change addresses the API docs build regression by reducing repeated work during API page generation and trimming a couple of broad build inputs.

- memoize API collection loads and sidebar generation in production builds
- scope C# and TypeScript API sidebars to the current package/module instead of rendering the full API catalog on every page
- apply a few sidebar micro-optimizations to avoid extra scans/sorts on scoped sidebars
- make `write-git-env.mjs` deterministic and avoid rewriting `.env.local` when values do not change
- replace the eager docs-wide glob in `Include.astro` with targeted docs collection lookup

## Validation

- did **not** run `pnpm build` locally
- `git diff --check`
- `pnpm exec prettier --check` on changed files
- isolated smoke test for `src/frontend/scripts/write-git-env.mjs`

## Notes

Route fan-out reduction was intentionally left out of this PR.

### DIFF
--- a/src/frontend/src/components/Include.astro
+++ b/src/frontend/src/components/Include.astro
@@ -1,6 +1,6 @@
 ---
 // Include a Markdown/MDX file from the docs collection inline.
-// Usage in MDX: <Include relativePath="reference/cli/includes/project-search-logic.description.md" />
+// Usage in MDX: <Include relativePath="reference/cli/includes/project-search-logic-description.md" />
 
 import { getEntry, render } from 'astro:content';
 


src/frontend/src/components/Include.astro
```diff
@@ -9,22 +11,17 @@ interface Props {
...
-const mdModules = {
-  ...(import.meta.glob('/src/content/docs/**/*.md', { eager: true }) as Record<string, unknown>),
-  ...(import.meta.glob('/src/content/docs/**/*.mdx', { eager: true }) as Record<string, unknown>),
-};
-
-// Normalize the provided path to match the glob keys (absolute from project root).
-const normalized = `/src/content/docs/${relativePath.replace(/^\/+/, '')}`;
-const mod = mdModules[normalized] as any;
+// Resolve the include through the docs collection instead of eagerly globbing
+// the entire docs tree into every page that uses <Include>.
```
eerhardt, line 15: What does this comment mean? Why is it saying "instead of eagerly globbing the entire docs tree"?

src/frontend/src/utils/packages.ts
```diff
@@ -4,11 +4,20 @@
 
 import { getCollection } from 'astro:content';
 
+let packagesPromise: Promise<any[]> | undefined;
+const shouldCachePackages = import.meta.env.PROD;
```
eerhardt, line 8: Why change the behavior on PROD?

src/frontend/src/utils/ts-modules.ts
```diff
@@ -4,11 +4,20 @@
 
 import { getCollection } from 'astro:content';
 
+let tsModulesPromise: Promise<any[]> | undefined;
+const shouldCacheTsModules = import.meta.env.PROD;
```
eerhardt, line 8: Same question on PROD

src/frontend/src/utils/ts-modules.ts
```diff
@@ -178,10 +203,7 @@ export function simplifyType(typeRef: string): string {
 
   // Clean assembly metadata from generic type arguments:
   // System.IEquatable`1[[TypeName, Assembly, Version=..., ...]] → System.IEquatable`1[[TypeName]]
-  stripped = stripped.replace(
-    /\[\[([^\],]+),\s*[^\]]*\]\]/g,
-    '[[$1]]',
-  );
+  stripped = stripped.replace(/\[\[([^\],]+),\s*[^\]]*\]\]/g, '[[$1]]');
```
eerhardt, line 206: (nit) was it random to decide when to put a single statement on a line vs. when to put it all on one line?

These 2 changes appear to be just formatting changes, but in the opposite directions.

src/frontend/src/components/Include.astro
IEvangelist: The code before, was doing this:

```
 ...(import.meta.glob('/src/content/docs/**/*.md', { eager: true }) as Record<string, unknown>),
  ...(import.meta.glob('/src/content/docs/**/*.mdx', { eager: true }) as Record<string, unknown>),
```

Which eagerly enumerated ALL matching .md and .mdx files. Which was not good for perf.

src/frontend/src/utils/packages.ts
IEvangelist: Locally, during dev, if we have a cached version - it's less of a concern. On PROD, we want to ensure we build from the latest regardless.

src/frontend/src/utils/ts-modules.ts
IEvangelist: Same as above.

src/frontend/src/utils/ts-modules.ts
IEvangelist: That's a good callout. We do have a linter/formatting rule. We don't enforce though, as it's a bit heavy handed - while it might add a bit of extra review work, we don't want to block PRs.

PR #605, opened 2026-03-24
Add ResourceName attribute to AddMailDev method

Meir017:
should probably task copilot to do this everywhere :)

### DIFF
--- a/src/frontend/src/content/docs/integrations/custom-integrations/hosting-integrations.mdx
+++ b/src/frontend/src/content/docs/integrations/custom-integrations/hosting-integrations.mdx
@@ -200,7 +200,7 @@ Replace the contents of the `Class1.cs` file in the `MailDev.Hosting` project, a
 // an alternative namespace.
 namespace Aspire.Hosting.ApplicationModel;
 
-public sealed class MailDevResource(string name) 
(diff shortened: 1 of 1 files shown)


src/frontend/src/content/docs/integrations/custom-integrations/hosting-integrations.mdx
eerhardt:
```suggestion
public sealed class MailDevResource([ResourceName] string name) 
```

Meir017: @eerhardt this pattern seems to be less common in the aspire repo - https://github.com/search?q=repo%3Amicrosoft%2Faspire%20%22(%5BResourceName%5D%22&type=code only 3 matches, compared with the extension method that adds a resource https://github.com/search?q=repo%3Amicrosoft%2Faspire+%22%5BResourceName%5D+string+name%22&type=code 60+ matches

eerhardt: If we did the analysis and attributing correctly it, the analyzer would catch all the places that didn't have this attribute but it passed an unattributed `string` into the base `Resource` ctor.

PR #609, opened 2026-03-25
Add TypeScript AppHost support to PostgreSQL integration docs

davidfowl:
Add TypeScript AppHost support to the PostgreSQL integration documentation.

Cherry-picked from `ts-integrations` branch (replaces #581, rebased onto `main`).


src/frontend/src/content/docs/integrations/databases/postgres/postgres-client.mdx
eerhardt, line 1: Should we rename this file because it is explicitly for .NET

src/frontend/src/content/docs/integrations/databases/postgres/postgres-connect.mdx
```diff
@@ -0,0 +1,228 @@
...
+    pass
+```
+
+</TabItem>
+<TabItem label="Go">
+
+Use the standard `pgx` driver:
+
+```bash title="Terminal"
+go get github.com/jackc/pgx/v5
```
eerhardt, line 137: Why this one and not https://github.com/lib/pq?

src/frontend/src/content/docs/integrations/databases/postgres/postgres-connect.mdx
```diff
@@ -0,0 +1,228 @@
...
+builder.Build().Run();
+```
+</TabItem>
+<TabItem label="TypeScript">
+```typescript title="TypeScript — apphost.ts"
+import { createBuilder } from './.modules/aspire.js';
+
+const builder = await createBuilder();
+
+const userName = await builder.addParameter("username", { secret: true });
```
eerhardt: Why does this one use parameters, and the C# one doesn't?

src/frontend/src/components/SyncedPivotSelector.astro
eerhardt, line 1: Can we throw some docs on this to explain what it is vs. the existing PivotSelector?

src/frontend/src/content/docs/integrations/databases/postgres/postgres-get-started.mdx
```diff
@@ -24,238 +21,315 @@ import postgresIcon from '@assets/icons/postgresql-icon.png';
...
+This updates your `aspire.config.json` with the PostgreSQL hosting integration package:
 
-var exampleProject = builder.AddUvicornApp("api", "./api", "main:app")
-                            .WithExternalHttpEndpoints()
-                            .WaitFor(postgresdb)
-                            .WithReference(postgresdb);
+```json title="aspire.config.json" ins={3}
+{
+  "packages": {
+    "Aspire.Hosting.PostgreSQL": "13.2.0"
```
eerhardt: Everywhere else uses `*` or some sort of version-agnostic string.

src/frontend/src/content/docs/integrations/databases/postgres/postgres-get-started.mdx
```diff
@@ -24,238 +21,315 @@ import postgresIcon from '@assets/icons/postgresql-icon.png';
...
+import psycopg
 ```
 
-</Pivot>
+Use `os.getenv()` to read the injected PostgreSQL properties:
 
-<Pivot id="javascript">
+```python title="Python - Obtain configuration properties"
+postgres_uri = os.getenv("POSTGRESDB_URI")
+postgres_name = os.getenv("POSTGRESDB_DATABASE", "db")
```
eerhardt: Why does this have a default value of `db`?

src/frontend/src/content/docs/integrations/databases/postgres/postgres-get-started.mdx
```diff
@@ -24,238 +21,315 @@ import postgresIcon from '@assets/icons/postgresql-icon.png';
...
+    .WithReference(postgresdb);
+```
 
 <Aside type="tip">
-    The full set of properties that Aspire injects depends on whether you passed a PostgreSQL server or database resource. For more information, see [Properties of the PostgreSQL resources](../postgres-client/#properties-of-the-postgresql-resources).
+    For the full C# AppHost PostgreSQL API surface, see [PostgreSQL Hosting integration](../postgres-host/).
 </Aside>
 
-### Use PostgreSQL resources in client code
+### Choose a consuming language for the C# AppHost quickstart
```
eerhardt: What does this title mean?

src/frontend/src/content/docs/integrations/databases/postgres/postgres-get-started.mdx
```diff
@@ -24,238 +21,315 @@ import postgresIcon from '@assets/icons/postgresql-icon.png';
...
-string jdbcconnectionstring = builder.Configuration.GetValue<string>("POSTGRESDB_JDBCCONNECTIONSTRING");
-string databasename = builder.Configuration.GetValue<string>("POSTGRESDB_DATABASE");
-```
+Use this path when your Aspire solution is modeled in `AppHost.cs`.
 
-</Pivot>
+### Add the PostgreSQL Hosting integration to the AppHost
 
-<Pivot id="python">
+```bash title="Terminal"
```
eerhardt, line 196: (nit) this looks different than the way we say to add the package to the app. Here:

Image

Lower in the page it is a selector:

Image

src/frontend/src/components/SyncedPivotSelector.astro
IEvangelist, line 1: This isn't needed, the existing PivotSelector is already persistent across articles.

IEvangelist (commented):
This feels broader than needed for adding TypeScript AppHost coverage to the PostgreSQL docs. I'd prefer to keep this PR focused on the docs changes, reuse the existing PivotSelector/Pivot pattern with one shared key across the Postgres articles so the AppHost choice persists, and split the unrelated sidebar behavior change into a separate PR. If we keep or add an in-page PivotSelector, please use `marginTop={1}` so the spacing matches the rest of the docs. For language examples, let's also make sure we're covering the canonical popular/actively maintained OSS clients, or explicitly explain why we're choosing one over another.

src/frontend/src/content/docs/integrations/databases/postgres/postgres-host.mdx
```diff
@@ -20,7 +19,19 @@ import postgresIcon from '@assets/icons/postgresql-icon.png';
 
 To get started with the Aspire PostgreSQL integrations, follow the [Get started with PostgreSQL integrations](/integrations/databases/postgres/postgres-get-started/) guide. If you want to learn how to use the PostgreSQL Entity Framework Core (EF Core) client integration, see [Get started with the PostgreSQL Entity Framework Core integrations](/integrations/databases/efcore/postgres/postgresql-get-started/).
 
-This article includes full details on the Aspire PostgreSQL Hosting integration, which allows you to model PostgreSQL server and database resources in your [`AppHost`](/get-started/app-host/) project.
+This article includes full details on the Aspire PostgreSQL Hosting integration, with examples for both `AppHost.cs` and `apphost.ts`, so you can model PostgreSQL server and database resources in your [`AppHost`](/get-started/app-host/) project.
+
+<SyncedPivotSelector
```
IEvangelist: I don't think we need a new wrapper component for this. I'd rather reuse PivotSelector directly here and let the shared key carry the selection across the related Postgres articles.

src/frontend/src/content/docs/integrations/databases/postgres/postgres-get-started.mdx
```diff
@@ -24,238 +21,315 @@ import postgresIcon from '@assets/icons/postgresql-icon.png';
...
-<InstallPackage packageName="Aspire.Hosting.PostgreSQL" shortName="postgresql" />
-
-Next, in the AppHost project, create instances of PostgreSQL server and database resources, then pass the database to the consuming client projects:
+Choose your AppHost language, then add PostgreSQL to the AppHost. For the full C# and TypeScript hosting API surface, see [PostgreSQL Hosting integration](../postgres-host/).
 
 <PivotSelector
-    title="Select your programming language to get started"
-    key="lang"
+    title="Select your AppHost language"
+    key="postgres-apphost-lang"
```
IEvangelist, line 36: If the goal is for the AppHost language choice to persist across the Postgres docs, this key should match the one used on postgres-host.mdx. Right now postgres-apphost-lang and postgres-host-apphost-lang won't behave like one shared selector across related articles.

src/frontend/src/components/starlight/Sidebar.astro
```diff
@@ -86,6 +86,19 @@ const referenceOverviewHref = `${base}${localePrefix}/reference/overview/`;
     </LinkButton>
   </div>
 </div>
+<script is:inline>
```
IEvangelist: This sidebar behavior change feels unrelated to adding TypeScript AppHost coverage for the PostgreSQL docs. Can we split it into a separate PR so this one stays focused?

src/frontend/src/content/docs/integrations/databases/postgres/postgres-connect.mdx
```diff
@@ -0,0 +1,228 @@
...
+    postgres_uri, autocommit=True
+) as conn:
+    # Use conn to query the database...
+    pass
+```
+
+</TabItem>
+<TabItem label="Go">
+
+Use the standard `pgx` driver:
```
IEvangelist: Can we either cover lib/pq here as well, or explain why we're standardizing on pgx? For these language-neutral connection pages, I'd like us to steer toward the most popular and actively maintained OSS clients, and be explicit about the recommendation.

src/frontend/src/content/docs/integrations/databases/postgres/postgres-get-started.mdx
```diff
@@ -24,238 +21,315 @@ import postgresIcon from '@assets/icons/postgresql-icon.png';
...
+```
 
-<Pivot id="python">
+Use `os.getenv()` to read the injected PostgreSQL properties:
 
-Use the information you have obtained about the PostgreSQL resource to connect to the database. Here is an example of how to connect using `psycopg`:
+```python title="Python - Obtain configuration properties"
+postgres_uri = os.getenv("POSTGRESDB_URI")
+postgres_name = os.getenv("POSTGRESDB_DATABASE", "db")
+```
```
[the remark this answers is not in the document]
davidfowl: Fixed — updated to `POSTGRESDB_DATABASENAME` (no default) to match the `DatabaseName` property convention.

src/frontend/src/content/docs/integrations/databases/postgres/postgres-connect.mdx
davidfowl: Added a note explaining pgx is recommended for new projects (better performance, native protocol support, active maintenance) with a link to `lib/pq` as an alternative.

src/frontend/src/content/docs/integrations/databases/postgres/postgres-get-started.mdx
davidfowl: Leaving as-is — hosting and client packages have genuinely different install flows (`aspire add` vs `dotnet add package`), so the different UX is expected.

src/frontend/src/content/docs/integrations/databases/postgres/postgres-client.mdx
davidfowl: Good call — will do this in a follow-up PR since renaming changes the URL slug and needs a redirect.

davidfowl:
@IEvangelist As we roll out TypeScript AppHost support to more integrations (Redis, MySQL, MongoDB, etc.), I'm noticing some patterns repeating across the Postgres pages that could be extracted into reusable components:

1. **`AppHostLanguageSwitcher`** — standardize the PivotSelector + key + "Select your AppHost language" boilerplate that every integration page will need
2. **`ClientLanguageTabs`** — the "install driver → import → read env vars → connect" pattern for JS/Python/Go/.NET that repeats across get-started and connect pages
3. **`ConnectionPropertyTable`** — env var property tables (Host, Port, Username, Password, Uri, etc.) that will repeat for every database integration

Postgres is the first integration with the C#/TS AppHost pivot, so it's a good time to extract these before we repeat the same patterns everywhere else. Do you agree this is worth doing, or should we keep things inline for now?

PR #621, opened 2026-03-26
Add CI gate check for conditional jobs

IEvangelist:
## Summary
- add a single always-run `ci-gate` job to fan in conditional CI results
- keep `frontend-build`, `apphost-build`, and `integration-test` conditional
- make it possible to require one stable status check in GitHub rulesets

## Why
Skipped jobs already complete successfully, but a single gate check makes branch protection configuration simpler and avoids requiring multiple conditional jobs directly.

After merging, update the repo ruleset to require `ci-gate` instead of the individual conditional checks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,3 +64,60 @@ jobs:
     uses: ./.github/workflows/integration.yml
     with:
       node_version:  '24.x'
+
+  ci-gate:
+    needs: [changes, frontend-build, apphost-build, integration-test]
+    if: ${{ always() && !cancelled() }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Verify CI results
+        shell: bash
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -64,3 +64,60 @@ jobs:
...
+          echo "apphost changed: $APPHOST_CHANGED"
+          echo "apphost-build result: $APPHOST_RESULT"
+          echo "integration-test result: $INTEGRATION_RESULT"
+
+          if [[ "$CHANGES_RESULT" != "success" ]]; then
+            echo "The changes job must succeed."
+            exit 1
+          fi
+
+          if [[ "$FRONTEND_CHANGED" == "true" ]]; then
```
adamint, line 95: is there not a more efficient way of structuring these?

PR #624, opened 2026-03-27
Refine CI gate workflow coverage

IEvangelist:
## Summary
- make the CI change detector treat workflow file changes as affecting both frontend and apphost validation
- include src/statichost in apphost change detection because AppHost builds StaticHost transitively
- remove the duplicate integration workflow and fold its dist/index.html assertion into frontend-build
- simplify ci-gate so it only validates the jobs that provide distinct coverage

## Why
The required ci-gate check could report success without exercising the workflows it depended on. A pull request that only changed files under .github/workflows could skip frontend-build and apphost-build while still passing ci-gate.

There was a second gap for src/statichost changes. AppHost references StaticHost directly, so those changes should be treated as apphost-affecting work.

The old integration workflow was also mostly duplicating frontend-build and apphost-build. It rebuilt both sides from source, did not consume artifacts from the other jobs, and did not perform a real smoke or integration check.

## Testing
- validated .github/workflows/ci.yml and .github/workflows/frontend-build.yml after the edits
- ran git diff --check to verify the final diff is clean

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,8 @@ jobs:
           base_sha="${{ github.event.pull_request.base.sha }}"
           head_sha="${{ github.event.pull_request.head.sha }}"
 
+          # If the workflow changed, build everything. if it hasn't,
+          # check for frontend and/or apphost changes...
           if git diff --quiet "$base_sha" "$head_sha" -- .github/workflows; then
             workflow=false
           else


.github/workflows/ci.yml
```diff
@@ -31,16 +31,30 @@ jobs:
             exit 0
           fi
 
-          if git diff --quiet "${{ github.event.pull_request.base.sha }}" "${{ github.event.pull_request.head.sha }}" -- src/frontend; then
-            frontend=false
+          base_sha="${{ github.event.pull_request.base.sha }}"
+          head_sha="${{ github.event.pull_request.head.sha }}"
+
+          if git diff --quiet "$base_sha" "$head_sha" -- .github/workflows; then
```
eerhardt, line 39: Can we add a comment explaining the logic here? "if the workflow changed, build everything. if it hasn't, check for frontend and/or apphost changes"

IEvangelist, line 39:
```suggestion
          # If the workflow changed, build everything. if it hasn't,
          # check for frontend and/or apphost changes...
          if git diff --quiet "$base_sha" "$head_sha" -- .github/workflows; then
```

PR #627, opened 2026-03-27
Fix astro-mermaid icon pack serialization

IEvangelist:
## Summary
- switch Mermaid icon packs to explicit URL-based configuration
- serve the custom Aspire icon pack from a static JSON asset
- align the frontend config with astro-mermaid's safe serialization requirements

## Testing
- pnpm --dir src/frontend build
  - confirmed the astro-mermaid icon pack serialization warning is gone
  - build then fails later on an unrelated Astro internal helpers export mismatch

Related to https://github.com/joesaby/astro-mermaid/issues/18


src/frontend/config/icon-packs.mjs
```diff
@@ -1,33 +1,21 @@
...
+// astro-mermaid now serializes icon packs as name/url pairs, so custom packs
+// must be served from a static JSON endpoint instead of an inline loader.
 
 export const iconPacks = [
   {
     // Search: https://icon-sets.iconify.design/logos/?keyword=svg+logos
     name: 'logos',
-    loader: () =>
-      fetch('https://unpkg.com/@iconify-json/logos@1/icons.json').then((res) => res.json()),
+    url: 'https://unpkg.com/@iconify-json/logos@1/icons.json',
```
eerhardt, line 9: Does this download this file during our build? and then we serve it?

or do we link to unpkg.com in our end site?

IEvangelist: > Does this download this file during our build? and then we serve it?

No, this is a client-side thing. They're lazy loaded.

> or do we link to unpkg.com in our end site?

Yes, in the browser network tab when you navigate to the http://localhost:4321/get-started/first-app-csharp-apphost/?lang=csharp page, for example, you'll see `icons.json` coming from unpkg.com and `aspire.json` being served locally for our custom icons.

eerhardt: Is that good? Do we want our site dependent on another site?

I guess if that is already how it happens today.... then no reason to be concerned with this PR.

Issue #230, opened 2026-01-09
At reflow resolution, header controls are not visible on the page.: A11y_Aspire website_Aspire page_Reflow


PoojaNamde:
### GitHub Tags:
#A11yTCS; #A11yMAS; #A11ySev2; #A11yWCAG; #DesktopWeb; #ChromiumEdge; #Win11; #Developer Community Partners; #SH_AspireWebsite_Web_Jan26; #WCAG1.4.10; #Reflow;

**"Please do not close this bug. This bug should only be closed by TCS, C+AI Accessibility tester after bug verification."**

### Environment Details:
Application Name: Aspire website
URL: https://aspire.dev/
Chromium Edge: Version 145.0.3775.0 (Official build) dev (64-bit)
Edition: Windows 11 Enterprise Insider Preview
Version: 24H2
OS build: 26540.1001

### Repro steps:

1. Hit the URL: https://aspire.dev/.
2. Observe the issue that at reflow resolution, header controls are not visible on the page.

### Actual Result:
When the page is set to reflow resolution, the header controls are no longer visible on the screen. Important interactive elements in the header appear to be hidden, making them inaccessible to users who rely on increased zoom levels.

### Expected Result:
All header controls should remain visible at reflow resolution without loss of content or functionality.

### User Impact:
Low‑vision users who rely on browser zoom may lose access to essential header controls, preventing them from navigating or performing key actions. This creates a significant usability barrier and impacts task completion.

### Attachment:

https://github.com/user-attachments/assets/670a353c-c6e2-400b-a4c6-1ef83a51948c

PR #666, opened 2026-04-02
Refactor AppHost docs and add a localized site tour

IEvangelist:
## Summary

Also fixes: #229 & #230

- consolidate the split AppHost documentation into shared guides with synced C# and TypeScript tabs, including the first-app, deploy-first-app, and add-existing-app flows
- add a full site tour experience with target highlighting, focus management, quick navigation, restart/resume behavior, overlay dismissal, and localized strings across all supported locales
- refresh the Starlight shell and related UI pieces, including the header, sidebar, footer preferences/socials, install CLI modal, edit link, page title, integration grid, and shared site styles
- move analytics assets into public scripts and update the tracking wiring
- update supporting docs, sidebar/topic configuration, breadcrumb routing, API reference pages, sample/stats scripts, and related guidance to align with the new AppHost and navigation work

## More detail
- Added `AppHostLangPivot.astro` and reworked documentation to reduce C#/TypeScript duplication while keeping AppHost-specific implementation details in tabs.
- Removed several split getting-started pages and merged them into unified scenario-based guides.
- Added `SiteTour.astro`, tour trigger/help affordances, footer preferences targeting, localized copy in `en.json` plus the non-English locale files, and the supporting header/sidebar/head integration changes.
- Moved `1ds.js` and `track.js` from page-level sources into `public/scripts/analytics/` and updated the site to use the new script locations.
- Included follow-on fixes for breadcrumbs and route normalization, sample repo references, contributor-guide markdown cleanup, sidebar topic data, redirects, API reference routing, and related documentation updates.

## Testing
- Validated changed files with editor diagnostics where applicable
- Did not run a full build or test pass

### DIFF
--- a/src/frontend/src/components/site-tour.ts
+++ b/src/frontend/src/components/site-tour.ts
@@ -313,10 +313,6 @@ export function isLandingPath(pathname: string): boolean {
   return parts.length === 1 && isLocaleSegment(parts[0]);
 }
 
-function isE2ETestRun(): boolean {
-  return document.head.querySelector('meta[name="e2e-tests"][content="true"]') !== null;
-}
-
 export function getStepHighlightPadding(stepId: string): number {
   if (stepId === 'search') {
     return 2;
@@ -481,12 +477,6 @@ class AspireSiteTour {
     }
 
     this.clearRefreshRetry();
-
-    if (!this.state.started && !this.hasSeenAnySteps() && !isE2ETestRun()) {
-      this.start(false);
-      return;
-    }
-
     this.clearPendingPaintLock();
 
     if (this.isOpen) {
@@ -917,6 +907,10 @@ class AspireSiteTour {
     this.singleStepMode = false;
     this.stepFilterMode = 'new';
     this.singleStepId = null;
+    if (!this.state.started) {
+      this.state.started = true;
+      this.saveState();
+    }
     this.openAt(0, true);
   }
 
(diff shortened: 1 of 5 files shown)


IEvangelist:
> I'm not a fan of some of the changes:
> 
> * I don't like moving top level items to a dropdown. It doesn't seem like a discoverable way to move between sections of docs. i.e. if you've never visited the docs site before, and you're looking at Fundamentals, and you want to learn about the dashboard, how do you know to click the dropdown to go to the dashboard section? I could imagine people land on the `/docs` page and think that all the docs that are available are in the tree.

>   However, I also don't like what we have today. So 🤷

Yeah, this one is tough. It's difficult architecting how to structure, organize, present, and expose all the content. This is the direction we're leaning now - but that might change.

> * I don't like that there are two search boxes. It makes people have to pick. Which means they need to figure out what each one is for. I think it should be removed along with the collapse sidebar button that I don't think people would find useful.

One is a site-wide search, and the other is a filter. They're different and that UX is common, github.com even has it for example. The collapsible sidebar is too:

image

> * I don't like tours at all. Maybe it's just me, but I immediately exit out of them and am annoyed the website took over the browser window. I don't think aspire.dev is complicated enough to require a tour.

They're easy to dismiss, but helpful when you need them. Also, one thing is certain after watching replays of user studies, most readers do not know how to use the site...maybe it's just something that we take for granted?

PR #672, opened 2026-04-04
Fix frontend e2e sidebar and WCAG regression failures

### LINKED ISSUE #676: Homepage hero image gradient bad
Image

IEvangelist:
## Summary

Fixes #676

image

Also, adds tests!

This PR fixes the frontend end-to-end failures that were showing up in the sidebar regression test and the WCAG AA audit suite, and it also folds in the broader frontend hardening work that landed alongside those fixes.

## What changed
- stabilized the Playwright UI regression and WCAG suites, including the API sidebar collapse-state persistence flow on reload
- hardened the shared accessibility helper so generated Starlight and Expressive Code blocks get stable labels, and made horizontally scrollable tables keyboard-focusable with generated labels where needed
- converted the remaining authored frontend JavaScript maintenance scripts and related authored surfaces to TypeScript, including the RSS route and shared utilities
- added broader frontend regression coverage: unit and end-to-end tests for analytics script delivery, cookie consent and preferences, RSS behavior, pivot behavior, and other UI regressions
- added custom Astro component coverage with both runtime render tests and compile-time prop-contract tests so component props remain discoverable and typed for consumers
- refactored heavier Astro frontmatter logic into shared TypeScript helpers, including the SimpleAppHostCode and floating icon layout logic, and normalized component prop typing across custom components
- cleaned up the Playwright package and script setup by removing the redundant CLI package and duplicate browser install path while preserving a dedicated local install helper
- updated the install CLI and VS Code extension docs so repeated code examples have distinct titles

## Why
This work addressed a few related frontend maintenance issues at the same time:
- the sidebar persistence test could race the post-reload UI state and intermittently interact with the wrong control before the page was ready
- Axe was flagging duplicate landmark names on generated code block regions, especially on tablet and mobile layouts where OS-aware tabs can promote different code blocks into landmarks after hydration
- mobile audit runs were also hitting scrollable table keyboard-access issues
- several authored frontend scripts and utilities still relied on JavaScript, which limited type safety and made larger refactors riskier than necessary
- the custom component surface needed stronger regression coverage so prop contracts and render behavior would stay stable as the frontend evolved

## Validation
Ran targeted validation against the Astro dev server and frontend test suites:

```text
pnpm --dir "e:\GitHub\aspire.dev\src\frontend" run test:unit:components
pnpm --dir "e:\GitHub\aspire.dev\src\frontend" exec playwright test tests/e2e/wcag-aa.spec.ts --project=desktop-chromium --project=tablet-chromium --project=mobile-chromium
pnpm --dir "e:\GitHub\aspire.dev\src\frontend" exec playwright test tests/e2e/rss.spec.ts
pnpm --dir "e:\GitHub\aspire.dev\src\frontend" exec playwright test tests/e2e/ui-regressions.spec.ts --project=desktop-chromium --grep "API sidebar collapse state persists across reloads" --repeat-each=3
```

Notes:
- the 2 skipped tests from the earlier validation run were the tablet and mobile variants of `API sidebar collapse state persists across reloads`
- that scenario intentionally skips on narrow viewports because the collapse and expand affordance only exists on wide desktop layouts

### DIFF
--- a/src/frontend/package.json
+++ b/src/frontend/package.json
@@ -34,7 +34,7 @@
     "test:e2e": "playwright test",
     "test:e2e:install": "playwright install chromium",
     "test:e2e:serve": "pnpm git-env && pnpm check-data && astro dev --host 127.0.0.1 --port 4321",
-    "lint": "eslint . --max-warnings 0",
+    "lint": "pnpm git-env && pnpm exec astro sync && eslint . --max-warnings 0",
     "format": "prettier -w --cache --plugin prettier-plugin-astro .",
     "update:all": "pnpm update:integrations && pnpm update:github-stats && pnpm update:samples",
     "update:integrations": "tsx ./scripts/update-integrations.ts",
--- a/src/frontend/src/content.config.ts
+++ b/src/frontend/src/content.config.ts
@@ -1,5 +1,3 @@
-/* Required in CI/CD, please leave for now:
-eslint-disable @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-member-access */
 import { defineCollection } from 'astro:content';
 import { docsLoader, i18nLoader } from '@astrojs/starlight/loaders';
 import { docsSchema, i18nSchema } from '@astrojs/starlight/schema';


src/frontend/src/components/AccessibleCodeButtons.astro
```diff
@@ -69,13 +175,14 @@
...
+    document.addEventListener('DOMContentLoaded', addAccessibilityEnhancements);
   } else {
-    addAccessibleLabels();
+    addAccessibilityEnhancements();
   }
 
   // Re-run when navigating with View Transitions
-  document.addEventListener('astro:page-load', addAccessibleLabels);
+  document.addEventListener('astro:page-load', addAccessibilityEnhancements);
+  window.addEventListener('resize', addAccessibilityEnhancements);
```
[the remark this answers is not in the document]
IEvangelist: Addressed in the current branch. The accessibility pass no longer runs directly on every resize event: it now goes through scheduleAccessibilityEnhancements(), which batches work through requestAnimationFrame, and each run queries the heading list once and reuses it when labeling code regions and scrollable tables.

.github/workflows/frontend-build.yml
```diff
@@ -37,6 +37,53 @@ jobs:
...
+        run: pnpm test:unit
+
+      - name: Install Playwright Chromium
+        run: pnpm exec playwright install --with-deps chromium
+
+      - name: Run frontend smoke tests
+        env:
+          ASTRO_TELEMETRY_DISABLED: 1
+        run: pnpm test:e2e
+
```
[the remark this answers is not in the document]
IEvangelist: Addressed. src/frontend/package.json now keeps `test:e2e` as plain `playwright test`, while `test:e2e:install` remains the one-time local helper. CI still installs Chromium explicitly in the workflow, so the duplicate install path is gone.

src/frontend/tests/e2e/wcag-aa.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+    await dismissCookieConsentIfVisible(page);
+
+    const results = await new AxeBuilder({ page })
+      .withTags(['wcag2a', 'wcag2aa'])
+      .options({
+        rules: {
+          'color-contrast': { enabled: true },
+        },
+      })
+      .analyze();
```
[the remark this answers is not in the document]
IEvangelist: Addressed. The WCAG spec now waits for waitForAccessibilityEnhancements(page) before running Axe, using the readiness marker exposed by AccessibleCodeButtons.astro so the audit runs after the client-side labeling pass has applied.

src/frontend/scripts/update-github-stats.ts
```diff
@@ -0,0 +1,94 @@
+import fs from 'fs';
+
+import { fetchWithProxy as fetch } from './fetch-with-proxy';
+
+const REPOS = [
+  'microsoft/aspire',
+  'dotnet/aspire-samples',
```
IEvangelist:
```suggestion
  'microsoft/aspire-samples',
```

src/frontend/src/components/FooterLinks.astro
```diff
@@ -3,87 +3,89 @@ import { Icon } from '@astrojs/starlight/components';
 import config from 'virtual:starlight/user-config';
 
 const links = config.social || [];
-const gitHubLink = links.filter(l => l.label == 'GitHub')[0].href || null;
-const discordLink = links.filter(l => l.label == 'Discord')[0].href || null;
-const is404 = Astro.locals.starlightRoute?.entry?.slug === "404";
+const gitHubLink = links.find((link) => link.label === 'GitHub')?.href ?? null;
```
JamesNK, line 6: This breaks 404 detection for localized pages. The old check `Astro.locals.starlightRoute?.entry?.slug === "404"` matched all locales (e.g. `/ja/404/`, `/zh-CN/404/`), but this pathname check only matches the English variants. Community links will incorrectly render on localized 404 pages.

Consider restoring the slug-based check, or using a regex/`includes` match on the pathname.

src/frontend/src/components/OsAwareTabs.astro
```diff
@@ -21,6 +21,31 @@ const { syncKey = 'terminal' } = Astro.props;
...
+
+          if (tabIndex < 0) return;
+
+          tabs.forEach((tab, index) => {
+            if (!(tab instanceof HTMLElement)) return;
+            const isSelected = index === tabIndex;
+            tab.setAttribute('aria-selected', isSelected ? 'true' : 'false');
+            tab.setAttribute('tabindex', isSelected ? '0' : '-1');
+          });
+
```
JamesNK, line 38: The `findIndex` check uses `instanceof HTMLAnchorElement`, but if Starlight renders tabs as `<button>` elements this will silently return -1 and `activateTabByLabel` will exit without switching any tab. The script then sets the localStorage preference at the bottom but the UI won't visually reflect it on first load.

Consider relaxing the check to `tab instanceof HTMLElement` (or just check `tab.textContent`) to handle both `<a>` and `<button>` tab implementations.

src/frontend/eslint.config.mjs
```diff
@@ -49,6 +49,10 @@ export default [
     files: ['**/*.js', '**/*.cjs', '**/*.mjs'],
     ...tseslint.configs.disableTypeChecked,
   },
+  {
+    files: ['tests/**/*.ts'],
+    ...tseslint.configs.disableTypeChecked,
```
JamesNK, line 54: This disables type-checked lint rules for **all** test files including `tests/typecheck/component-props.contracts.ts`, whose entire purpose is compile-time type validation via `@ts-expect-error` annotations. Consider narrowing the glob to just `tests/e2e/**/*.ts` if only the Playwright files are causing issues.

src/frontend/src/components/api-reference/MemberCard.astro
```diff
@@ -131,192 +154,275 @@ if (formattedSigRaw && (member.kind === 'property' || member.kind === 'indexer')
...
-        )}
-      </div>
-    </aside>
-  )}
+  {
+    isExperimental && (
+      <aside class="api-experimental-aside">
+        <svg
+          width="16"
+          height="16"
```
JamesNK, line 186: Nit: the section labels (Parameters, Returns, Exceptions, Remarks, Examples) were demoted from `<h5>` to `<p>`. These are no longer in the heading hierarchy, which makes it harder for screen-reader users to navigate within a member card by headings. Since this file is one of the most heading-dense surfaces on the API reference, the `<h5>` elements were useful navigation landmarks.

Was this intentional (e.g. to fix an axe violation about heading order), or incidental?

src/frontend/scripts/fetch-with-proxy.ts
```diff
@@ -0,0 +1,31 @@
...
+}
+
+export function fetchWithProxy(url: string | URL, options: RequestInit = {}): Promise<Response> {
+  if ((!httpsProxyAgent && !httpProxyAgent) || options.agent) {
+    return fetch(url, options);
+  }
+
+  const parsedUrl = url instanceof URL ? url : new URL(url);
+  return fetch(url, { ...options, agent: selectAgent(parsedUrl) });
+}
```
JamesNK, line 31: Nit: missing trailing newline at end of file.

src/frontend/src/components/api-reference/MemberCard.astro
IEvangelist: Intentional, they should not have been headings before.

PR #705, opened 2026-04-12
Add environments documentation guide

davidfowl:
## Summary

Adds deployment docs for Aspire environments and updates related deployment guides so environment-specific behavior is easier to discover.

### New page: /deployment/environments/

- Explains the Aspire environment as an AppHost concept
- Shows how to pass environment names to the AppHost for local start and deployment commands
- Shows how to read the environment in C# and TypeScript AppHosts
- Covers common patterns for:
  - setting runtime environment variables on child resources
  - parameter resolution and per-environment values
  - using execution context for run vs publish decisions
  - mapping GitHub Environments to Aspire environments

### Updated: /deployment/azure/azure-developer-cli/

- Clarifies that azd env select does not automatically set DOTNET_ENVIRONMENT
- Clarifies the difference between Aspire parameters and azd infrastructure inputs

### Updated: /deployment/docker-compose/

- Adds multi-environment deployment guidance
- Adds a command-reference link for aspire do prepare-* steps

### Sidebar

- Adds Environments to the Deployment section

PR #701, opened 2026-04-10
Add FAQ page and homepage FAQ link

BethMassi:
## Summary
- add a new getting started FAQ page
- add the FAQ to the docs sidebar
- link to the FAQ from the homepage

Fixes #690

PR #708, opened 2026-04-13
fix: fetch PR head history for Lunaria builds

IEvangelist:
## Summary

Follow up to #707 by fetching the PR head ref in the reusable frontend workflow before the Lunaria build.

## Why

The earlier `git fetch --unshallow` change was necessary, but not sufficient for all pull request checkouts. We still have failing `frontend-build` runs where Lunaria logs:

```text
[info] Shallow repository detected. A clone of your repository's history will be downloaded and used.
```

and then fails on a newly added docs page such as:

- `/src/frontend/src/content/docs/get-started/faq.mdx` in #701
- `/src/frontend/src/content/docs/deployment/environments.mdx` in #705

## Root cause

On some PR merge checkouts (`refs/pull/<n>/merge`), Lunaria still sees the repository as shallow and switches into its fallback history path. That path cannot resolve git history for docs pages that only exist on the PR branch.

## Fix

When this reusable workflow is running for a pull request, fetch:

```bash
refs/pull/<n>/head
```

explicitly, with `--unshallow`, before the frontend build. This makes the PR head history available even when the checkout was created from the merge ref.

For non-PR builds, keep the existing `git fetch --unshallow` fallback.

## Validation

Simulated PR merge checkouts for the known failing cases:

- PR #701: shallow `true` -> `false`, `faq.mdx` history restored
- PR #705: shallow `true` -> `false`, `environments.mdx` history restored

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,6 +68,7 @@ jobs:
     uses: ./.github/workflows/frontend-build.yml
     with:
       node_version: '24.x'
+      pull_number: ${{ github.event_name == 'pull_request' && github.event.pull_request.number || '' }}
 
   apphost-build:
     needs: changes
--- a/.github/workflows/frontend-build.yml
+++ b/.github/workflows/frontend-build.yml
@@ -34,13 +34,22 @@ jobs:
           PULL_NUMBER: ${{ inputs.pull_number }}
         run: |
           if [ -n "$PULL_NUMBER" ]; then
-            git fetch origin \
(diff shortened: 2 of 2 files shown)


IEvangelist:
Addressed the review feedback in the latest push:

- validate `pull_number` as numeric before using it in the refspec
- stop swallowing real `git fetch` failures
- use a clearer `refs/remotes/origin/pull/<n>/head` destination ref and store the refspec in a variable

The guarded fetch path still reproduces correctly against the known failing PR shapes (#701 and #705): shallow `true` -> `false`, and the new docs page history is available after the fetch.

davidfowl (approved):

PR #720, opened 2026-04-15
Improve deployment overview and CI/CD guidance

davidfowl:
## Summary
- add a workflow-first CI/CD overview page and reorganize the deployment navigation around the pipeline model, CI/CD, and targets
- rewrite the deployment overview around the pipeline model, clarify publish vs deploy, and explain compute environments, parameters, and environments together
- refresh the deployment landing page, pipelines overview, and example lifecycle workflow to match the new terminology and guidance
- refine deployment nav labeling, expand CI/CD on first mention, and style hero inline code to match the topic accent
- update the doc-writer skill to require visible C# and TypeScript AppHost examples

## Validation
- pnpm --dir ./src/frontend run lint
- pnpm --dir ./src/frontend run test:unit:docs
- pnpm --dir ./src/frontend run test:unit:components


src/frontend/config/sidebar/deployment.topics.ts
```diff
@@ -68,56 +68,88 @@ export const deploymentTopics: StarlightSidebarTopicsUserConfig = {
...
     {
-      label: 'Deploy to Kubernetes',
+      label: 'Kubernetes',
       slug: 'deployment/kubernetes',
     },
     {
-      label: 'Deploy to Azure',
-      collapsed: false,
+      label: 'Azure',
       translations: {
```
IEvangelist: If it's a product name that shouldn't be translated, we can omit the translations node altogether, in other words, when it's the same regardless of locale - omit the translations.

src/frontend/src/content/docs/deployment/ci-cd.mdx
```diff
@@ -0,0 +1,183 @@
+---
+title: CI/CD overview
+description: Use the Aspire CLI as the source of truth in CI/CD and design workflows around environments, parameters, and target-specific deployment guidance.
+---
+
+import { CardGrid, LinkCard, Steps } from '@astrojs/starlight/components';
+import LearnMore from '@components/LearnMore.astro';
+
+CI/CD with Aspire has two layers that work together:
```
IEvangelist:
```suggestion
<abbr title='Continuous Integration & Continuous Delivery' data-tooltip-placement='top'>CI/CD</abbr> with Aspire has two layers that work together:
```

src/frontend/src/components/TopicHero.astro
```diff
@@ -54,7 +62,13 @@ const floatingIconConfigs = buildFloatingIconLayout(title, floatingIcons);
       {title}{' '}
       {highlight && <span class="topic-hero-highlight">{highlight}</span>}
     </h1>
-    <p class="topic-hero-subtitle">{subtitle}</p>
+    <p class="topic-hero-subtitle">
+      {
+        subtitleSegments.map((segment) =>
+          segment.kind === 'code' ? <code>{segment.text}</code> : segment.text
```
IEvangelist, line 68: Nice! I think since we're rendering this as code now in the hero, we should use the accent color for the hero card to apply some custom inline code effect to make it stand out a bit more:

Image

Right now, it only seems to apply the monospaced font, rather than bg color or text color. I think we could improve that still.

IEvangelist (commented):
I think we need to change the top-level sidebar name from "Pipelines" to something else. Right now, it reads like Azure Pipelines, or that's where my head went, specially seeing CI/CD at the same level. It's almost like we need a different name there.

Also, since you're making these changes could you please fix the pipelines overview doc? It has an `overview` heading that should be renamed, as we get an overview in the TOC for free.

Image

Also, it uses pivots and it should instead use Tabs/TabItem for the apphost languages. And in the empty tab for TS, we should have a note to the user explaining that TS doesn't have those APIs supported yet.

davidfowl:
@IEvangelist Addressed in 7be90b1. I renamed the sidebar group to **Pipeline model** to avoid the Azure Pipelines/CI-CD ambiguity, renamed the pipelines page's redundant `Overview` heading, and replaced the AppHost language pivots with `Tabs`/`TabItem`. I also added the TypeScript note explaining that the custom `builder.pipeline` APIs aren't supported there yet. If you'd prefer **Aspire pipeline** over **Pipeline model**, I can make that follow-up too.

IEvangelist (approved):

PR #740, opened 2026-04-21
Harden twoslash config against silent regressions

slang25:
Two small defensive changes in `ec.config.mjs`, layered on top of #741.

## Changes

- **Throw instead of warn when `src/data/twoslash/aspire.d.ts` is missing.** Since #741 source-controls the bundle, a missing file means the working tree is corrupted — not a normal path. Catching that at build time beats silently shipping samples that can't resolve `./.modules/aspire.js`.
- **Flip `noErrorValidation` to `false`** so unannotated TS errors fail the build rather than rendering as squigglies in the shipped HTML. Samples that deliberately illustrate a compiler error can opt in with `// @errors: <codes>`; otherwise an error means the sample (or the generated SDK shape) is wrong and should be fixed, not shown to readers.

Originally this PR also carried a `frontend.esproj` publish-time MSBuild target. That fix is no longer needed: #741 source-controls the generated bundle, so `dotnet publish` no longer depends on a pre-build type-generation step. Dropped it.

## Test plan

- [x] Full `pnpm build:skip-search` passes with `noErrorValidation: false` — no existing sample has an unannotated TS error
- [x] Generated HTML in `dist/` has zero `Cannot find module` / twoslash-error markers
- [x] Deleting `src/data/twoslash/aspire.d.ts` and re-importing `ec.config.mjs` throws; restoring it loads cleanly
- [x] `pnpm test:unit:twoslash-types` passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/frontend/ec.config.mjs
+++ b/src/frontend/ec.config.mjs
@@ -9,15 +9,17 @@ import { pluginDisableCopy } from './src/expressive-code-plugins/disable-copy.mj
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
 const ASPIRE_TYPES_PATH = resolve(__dirname, 'src/data/twoslash/aspire.d.ts');
-const aspireTypes = existsSync(ASPIRE_TYPES_PATH)
-  ? readFileSync(ASPIRE_TYPES_PATH, 'utf8')
-  : '';
 
-if (!aspireTypes) {
-  // Non-fatal — twoslash blocks that import the SDK will just show `any`.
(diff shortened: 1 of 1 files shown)


src/frontend/frontend.esproj
IEvangelist, line 1: These changes we will not accept, our build pipeline manages a lot of these already.

IEvangelist (approved):

IEvangelist:
Hey @slang25 is this still needed?

PR #754, opened 2026-04-27
Fix localized API docs routing

IEvangelist:
## Summary

This draft PR fixes localized API reference URLs that were being exposed to users and indexed by Google Search Console even though the API docs are not localized. When a visitor used the site in a non-English locale and then searched or browsed C# / TypeScript API reference content, the locale segment could carry forward into URLs such as `/{locale}/reference/api/...`. Those localized API URLs do not have corresponding pages, which produced 404s and a large volume of Search Console indexing failures.

## Changes

- Canonicalize localized API reference URLs back to the supported unlocalized `/reference/api/...` route with a 308 redirect.
- Share API reference route matching and locale path helpers across middleware, the early locale script, footer preferences, and unit tests.
- Keep API docs always unlocalized while preserving the visitor's preferred content locale so returning to non-API pages restores the selected locale.
- Update footer Resources/Legal organization while adding the external Aspire brand link.

## Validation

- `pnpm --dir src/frontend test:unit:ts-api`
- VS Code diagnostics for touched Astro/TypeScript files
- Local Playwright CLI UX checks for Italian/French locale flows, API search/navigation, canonical API URLs, and return-to-localized-site behavior

No full build was run.


.github/skills/doc-writer/SKILL.md
```diff
@@ -819,7 +825,7 @@ curl -sSL https://aspire.dev/install.sh | bash
 irm https://aspire.dev/install.ps1 | iex
 ```
 
-For complete installation instructions, see [Install Aspire CLI](/get-started/install-cli/).
+For complete installation instructions, see [Install Aspire CLI](https://aspire.dev/get-started/install-cli/).
```
davidfowl, line 828: Why this change?

adamint (approved):

.github/skills/doc-writer/SKILL.md
IEvangelist: Because skills don't resolve relative paths without extension. They can however make web calls, so it was either that or drop a full path to the .mdx file.

IEvangelist: I do plan on updating the docs-writer skill to better break down specific sections into references, as the SKILL.md is much larger than was originally intended.

PR #758, opened 2026-04-28
Live status header icon + redesigned videos page

IEvangelist:
# Live status header icon + redesigned videos page

Adds a real-time **live indicator** to the aspire.dev site header, immediately left of cookie preferences, and replaces the legacy `/community/videos/` curated list with a focused two-tab live page for YouTube and Twitch.

When a stream is live, state is pushed to connected clients over Server-Sent Events from StaticHost. The header icon strobes everywhere except `/community/videos/` and while native Picture-in-Picture is already open. Clicking the live header icon opens a native Document Picture-in-Picture window when supported, and falls back to the videos page otherwise. The PiP window is tracked as site-global state so it survives Astro client-side navigations without recreating the embed or interrupting playback; closing native PiP only closes PiP and does not redirect.

## What's included

- Frontend live icon, SSE client, header wiring, and videos page rewrite.
- Native Document Picture-in-Picture controller for live embeds, with YouTube/Twitch embed selection from the server-provided `primarySource` and PiP-open state shared across Astro navigations.
- YouTube/Twitch tab icons and offline channel embeds: YouTube loads the `aspiredotdev` channel live-stream embed and Twitch loads the `aspiredotdev` channel player even when no live event is active.
- ASP.NET Core live-status backend with `GET /api/live`, `GET /api/live/stream`, Twitch EventSub webhook, YouTube WebSub webhook, and a dev-only override endpoint.
- Twitch and YouTube background services with resilient named `HttpClient`s, webhook signature validation, reconciling polls/subscription renewal, and safe idle behavior when provider credentials are missing.
- AppHost dashboard URLs, Scalar API reference, and local dashboard commands for offline/online fake states and signed fake webhook invocations.
- Server-side protection for dev/dashboard commands via a per-run secret AppHost parameter sent as `X-Aspire-Live-Dev-Command-Key` and validated by StaticHost.
- `wwwroot/.gitignore` guard so local copies of the built frontend output cannot be accidentally added to source control.
- Frontend scripts `build:statichost` and `build:statichost:skip-search` that build Astro directly into `src/statichost/StaticHost/wwwroot` while preserving the Scalar assets and ignore guard.

## Local testing

Start the AppHost from the worktree root with `aspire start --isolated --apphost .\src\apphost\Aspire.Dev.AppHost\Aspire.Dev.AppHost.csproj`, open the Aspire dashboard, select the `aspiredev` resource, and run the live-status commands from the Actions context menu.

To refresh the local StaticHost static files without search indexing, run from `src/frontend`:

```powershell
pnpm build:statichost:skip-search
```

In AppHost run mode, StaticHost gets `Live__EnableDevEndpoint=true` and per-run secret parameters for the dev command key and fake webhook signing keys. Without real provider API credentials, Twitch and YouTube workers stay idle, so the feature is effectively off until a dashboard command triggers it.

Useful endpoints on the `aspiredev` resource: `/api/live`, `/api/live/stream`, `/api/live/twitch/webhook`, `/api/live/youtube/webhook`, `/api/live/_dev/set`, and `/scalar/v1`. The Twitch webhook is POST-only for notifications and the YouTube webhook supports GET verification plus POST notifications; plain GETs now return descriptive text instead of an unhelpful 404.

## Tests

- `tests/StaticHost.Tests` covers webhook HMAC verification, YouTube Atom parsing, JSON shape, live-state aggregation, coalescing, subscribe/unsubscribe, and timing behavior.
- `src/frontend/tests/unit/live-status.vitest.test.ts` covers the live-status client public API.
- `src/frontend/tests/e2e/live-status.spec.ts` covers mocked `/api/live` + SSE behavior, header strobing, videos-page no-strobe behavior, PiP-open strobe suppression, native PiP click interception, PiP state surviving Astro client-side navigation, and idle channel embeds.

## Configuration

```json
"Live": {
  "EnableDevEndpoint": false,
  "DevCommandSecret": "",
  "Twitch": { "ClientId": "", "ClientSecret": "", "WebhookSecret": "", "ChannelLogin": "aspiredotdev" },
  "YouTube": { "ApiKey": "", "WebhookSecret": "", "ChannelHandle": "@aspiredotdev" }
}
```

User-secrets in dev, env vars / Key Vault in prod. Missing provider secrets degrade safely: the workers log and remain idle, while the API/SSE endpoints continue serving the non-live state.


src/apphost/Aspire.Dev.AppHost/AppHost.cs
```diff
@@ -1,10 +1,121 @@
...
+using System.Text;
+
 var builder = DistributedApplication.CreateBuilder(args);
 
 var staticHostWebsite = builder.AddProject<Projects.StaticHost>("aspiredev")
     .WithExternalHttpEndpoints();
 
 if (builder.ExecutionContext.IsRunMode)
 {
+    var liveDevCommandSecret = builder.AddParameter("live-dev-command-secret", LiveDevCommands.NewSecret, secret: true);
```
eerhardt: Does this mean people can't `aspire run` the site without setting these parameters?

src/apphost/Aspire.Dev.AppHost/AppHost.cs
```diff
@@ -20,3 +131,141 @@
 }
 
 builder.Build().Run();
+
+internal static class LiveDevCommands
```
eerhardt: What do you think about putting all of this in a separate file, and even the WithUrls and WithHttpCommand calls above into a `LiveExtensions.cs` file?

src/apphost/Aspire.Dev.AppHost/AppHost.cs
```diff
@@ -1,10 +1,121 @@
...
+            }
+
+            ctx.Urls.Add(new() { Url = "/api/live", DisplayText = "Live status (JSON)", Endpoint = endpoint });
+            ctx.Urls.Add(new() { Url = "/api/live/stream", DisplayText = "Live status (SSE stream)", Endpoint = endpoint });
+            ctx.Urls.Add(new() { Url = "/api/live/twitch/webhook", DisplayText = "Twitch EventSub webhook (POST)", Endpoint = endpoint });
+            ctx.Urls.Add(new() { Url = "/api/live/youtube/webhook", DisplayText = "YouTube WebSub webhook (GET/POST)", Endpoint = endpoint });
+            ctx.Urls.Add(new() { Url = "/api/live/_dev/set", DisplayText = "Live dev override", Endpoint = endpoint });
+            ctx.Urls.Add(new() { Url = "/scalar/v1", DisplayText = "API reference (Scalar)", Endpoint = endpoint });
+        })
+        .WithHttpCommand(
```
eerhardt: Do we have tests that run these commands?

src/statichost/StaticHost/StaticHost.csproj
```diff
@@ -20,11 +20,28 @@
     <PackageReference Include="OpenTelemetry.Instrumentation.AspNetCore" Version="1.15.0" />
     <PackageReference Include="OpenTelemetry.Instrumentation.Http" Version="1.15.0" />
     <PackageReference Include="OpenTelemetry.Instrumentation.Runtime" Version="1.15.0" />
+    <PackageReference Include="Microsoft.Extensions.Http.Resilience" Version="9.10.0" />
```
eerhardt: Why so old? 10.5 is the latest.

src/frontend/src/components/live-status.ts
```diff
@@ -0,0 +1,263 @@
...
+}
+
+function onLivePipChange(evt: CustomEvent<{ open: boolean }>): void {
+  pipOpen = evt.detail.open;
+  syncDom();
+}
+
+async function seed(): Promise<void> {
+  try {
+    const res = await fetch('/api/live', { headers: { Accept: 'application/json' } });
```
adamint: How is this expected to work under `aspire run`? The AppHost exposes Vite and StaticHost as separate origins, but both this fetch and the `EventSource` below use Vite's `/api/live`, and there is no proxy in `astro.config.mjs`. I think we need to proxy `/api/live*` to the StaticHost endpoint or inject its URL; otherwise the live indicator never receives state from the normal local-dev URL.

src/statichost/StaticHost/Live/LiveEndpoints.cs
```diff
@@ -0,0 +1,387 @@
...
+
+        var topic = query["hub.topic"].ToString();
+        var expected = $"https://www.youtube.com/xml/feeds/videos.xml?channel_id={options.Value.YouTube.ChannelId}";
+        if (!string.IsNullOrEmpty(options.Value.YouTube.ChannelId) &&
+            !string.Equals(topic, expected, StringComparison.OrdinalIgnoreCase))
+        {
+            return Results.NotFound();
+        }
+
+        var challenge = query["hub.challenge"].ToString();
```
adamint: The current Front Door route caches `/*` with `IgnoreQueryString`, so this response can be cached and reused for a different `hub.challenge`. Could we bypass CDN caching for `/api/*` (or otherwise make this verification response uncacheable at Front Door)? WebSub renewals can fail when they receive the previous challenge.

src/statichost/StaticHost/Live/LiveEndpoints.cs
```diff
@@ -0,0 +1,387 @@
...
+            return Results.StatusCode(StatusCodes.Status503ServiceUnavailable);
+        }
+
+        var headers = context.Request.Headers;
+        var messageId = headers["Twitch-Eventsub-Message-Id"].ToString();
+        var timestamp = headers["Twitch-Eventsub-Message-Timestamp"].ToString();
+        var signature = headers["Twitch-Eventsub-Message-Signature"].ToString();
+        var messageType = headers["Twitch-Eventsub-Message-Type"].ToString();
+
+        if (string.IsNullOrEmpty(messageId) || string.IsNullOrEmpty(timestamp) || string.IsNullOrEmpty(signature))
```
adamint, line 171: Twitch signs the timestamp, but we also need to reject old messages. Without freshness validation, a captured callback becomes valid again after this 1,024-entry dedupe cache evicts it or the process restarts. Could we parse the header and reject messages outside Twitch's recommended window before handling the body?

src/statichost/StaticHost/Live/LiveEndpoints.cs
```diff
@@ -0,0 +1,387 @@
...
+        }
+
+        if (RequiresDevCommandSecret(env, liveOptions, twitch.IsConfigured) &&
+            !HasValidDevCommandSecret(context, liveOptions))
+        {
+            logger.LogWarning("Twitch dev webhook command rejected because the dev command secret header was missing or invalid.");
+            return Results.Unauthorized();
+        }
+
+        if (!s_twitchDedup.TryRegister(messageId))
```
adamint: This dedupe path also handles `webhook_callback_verification`. If Twitch retries the same verification because the first response was lost, we return an empty 200 instead of the challenge and the subscription fails. Could we dedupe notifications/revocations only, or cache and replay the verification response?

src/statichost/StaticHost/Live/LiveEndpoints.cs
```diff
@@ -0,0 +1,387 @@
...
+                return Results.Unauthorized();
+            }
+
+            var videoId = YouTubeWebhookHandler.ExtractVideoId(bodyBytes);
+            broadcaster.Update(new LiveStatusUpdate { YouTube = new YouTubeStatus(videoId is not null, videoId) });
+            logger.LogInformation("YouTube dev webhook accepted without API key; videoId={VideoId}", videoId);
+            return Results.Ok();
+        }
+
+        // Run the confirming poll out of band; we don't want to block the hub.
```
adamint: Each signed notification starts an untracked `Task.Run` with `CancellationToken.None`, so retries can fan out concurrent quota-consuming polls and keep running through shutdown or complete out of order. Could this go through a bounded/coalescing worker tied to host cancellation instead?

src/statichost/StaticHost/Live/LiveStatusBroadcaster.cs
```diff
@@ -0,0 +1,211 @@
...
+        var channel = Channel.CreateBounded<LiveStatus>(new BoundedChannelOptions(8)
+        {
+            FullMode = BoundedChannelFullMode.DropOldest,
+            SingleReader = true,
+            SingleWriter = false,
+        });
+
+        // Seed with current state so a freshly-connected client doesn't have to
+        // wait for the next change to render correctly.
+        channel.Writer.TryWrite(Current);
```
adamint: There is a race between this seed and adding the writer under `_gate`: a flush in between sends the new state before this subscriber exists, leaving it with the stale seed until another update. Could we take the snapshot and register the writer atomically under the same lock?

src/statichost/StaticHost/Live/LiveStatusBroadcaster.cs
```diff
@@ -0,0 +1,211 @@
...
+    {
+        lock (_gate)
+        {
+            var basis = _pending ?? _current;
+            var twitch = update.Twitch ?? basis.Twitch;
+            var youtube = update.YouTube ?? basis.YouTube;
+
+            // Sticky primary: keep the previous primary if it's still live;
+            // otherwise pick whichever source is live.
+            var primary = ResolvePrimary(_current.PrimarySource, twitch, youtube);
```
adamint: Should this use `basis.PrimarySource`? During the coalescing window, a YouTube update can establish pending state and then a Twitch update still reads `_current` (offline), causing Twitch to win even though YouTube arrived first. That breaks the documented sticky-first-source behavior.

src/statichost/StaticHost/Live/LiveStatusBroadcaster.cs
```diff
@@ -0,0 +1,211 @@
...
+            var isLive = twitch.Live || youtube.Live;
+            var now = _time.GetUtcNow();
```
adamint, line 110: `now` becomes part of the record before the no-op comparison below, so two otherwise identical reconciliations are never equal and always broadcast. Could we compare the substantive fields first and only update `UpdatedAt` when state actually changes?

src/frontend/src/components/LiveVideosTabs.astro
```diff
@@ -0,0 +1,118 @@
...
+    return document.querySelector<HTMLElement>(PAGE_SELECTOR);
+  }
+
+  function activateTab(label: 'YouTube' | 'Twitch'): void {
+    const root = findTabRoot();
+    if (!root) return;
+    const tabBtn = Array.from(
+      root.querySelectorAll<HTMLElement>('[role="tab"]'),
+    ).find((b) => b.textContent?.trim() === label);
+    tabBtn?.click();
```
adamint, line 56: This programmatic click triggers the listener below and sets `userSticky = true` on the first automatic selection. After that, later provider changes never switch tabs automatically. Could we distinguish programmatic activation from a user click?

src/frontend/src/components/live-status.ts
```diff
@@ -0,0 +1,263 @@
...
+  }
+  return () => listeners.delete(listener);
+}
+
+export function init(): void {
+  // Always re-sync the DOM so freshly swapped header buttons get the right state.
+  syncDom();
+  if (started) return;
+  started = true;
+  void seed();
```
adamint, line 255: `seed()` and the SSE connection start concurrently. If SSE delivers a newer snapshot before the fetch completes, the forced seed response can overwrite it with stale state until another event arrives. Could we ignore a seed after the first SSE state or order snapshots by `updatedAt`?

src/frontend/tests/e2e/live-status.spec.ts
```diff
@@ -0,0 +1,805 @@
...
+        },
+      });
+      await route.fulfill({
+        status: 200,
+        headers: {
+          'content-type': 'text/event-stream',
+          'cache-control': 'no-store',
+          connection: 'keep-alive',
+        },
+        body: body as unknown as Buffer,
```
adamint: Does this actually stream in Playwright? `route.fulfill` accepts a string or `Buffer`, so this cast passes a `ReadableStream` object at runtime rather than an SSE body. These tests can then succeed from the JSON seed without proving updates or reconnects. I think we should mock `EventSource` or run a small streaming endpoint.

src/frontend/scripts/build-static-host.mjs
```diff
@@ -0,0 +1,62 @@
...
+    },
+  );
+
+  restore(join(staticHostWwwroot, 'scalar'), 'scalar');
+  const gitignorePath = join(staticHostWwwroot, '.gitignore');
+  restore(gitignorePath, '.gitignore');
+  if (!existsSync(gitignorePath)) {
+    writeFileSync(gitignorePath, gitignoreContents);
+  }
+} finally {
```
adamint: If `astro build` throws after cleaning `wwwroot`, we skip both restore calls and immediately delete the only preserved copies in `finally`. Could we restore `scalar` and `.gitignore` from `finally` so a failed local build does not remove tracked assets?

src/statichost/StaticHost/Live/Twitch/TwitchEventSubService.cs
```diff
@@ -0,0 +1,142 @@
...
+        }
+        catch (OperationCanceledException) { return; }
+
+        while (!stoppingToken.IsCancellationRequested)
+        {
+            try
+            {
+                await ReconcileAsync(stoppingToken).ConfigureAwait(false);
+            }
+            catch (OperationCanceledException) { return; }
```
adamint: This catches every `OperationCanceledException`, including `HttpClient` timeouts where `stoppingToken` is not canceled, and permanently stops the worker after a transient cancellation. Could we only return when `stoppingToken.IsCancellationRequested` and let other cancellations follow the retry path? The YouTube worker has the same pattern.

adamint (approved):
I left inline comments for the must-fix local-dev, webhook validation/caching, broadcaster ordering, and frontend state/test issues. These all look fixable in place, and CI is green at this head. I could not run the site or browser paths inside the isolated worker because the dependencies and local server were unavailable there, so this pass relies on CI plus direct source inspection for that proof.

src/frontend/src/components/live-status.ts
IEvangelist: Good catch — under `aspire run` Vite and StaticHost are separate origins. The AppHost now injects StaticHost's origin as `ASPIRE_STATICHOST_URL`, and `astro.config.mjs` adds a dev-only Vite proxy for `/api/live` (which covers both this fetch and the `EventSource` stream). It's unset in CI/prod where StaticHost serves the site and the API from one origin, so the proxy is simply omitted there.

src/statichost/StaticHost/Live/LiveEndpoints.cs
IEvangelist: Agreed. Everything under `/api/live` is now marked `Cache-Control: no-store` centrally via a group endpoint filter, so Front Door can't cache or replay a snapshot or a verification challenge. Removed the redundant per-handler sets.

src/statichost/StaticHost/Live/LiveEndpoints.cs
IEvangelist: Added a freshness check: `TwitchWebhookHandler.IsFresh` rejects any notification whose signed `Message-Timestamp` is older than 10 minutes (with a 1-minute future-skew tolerance), so a captured-but-signed callback can't be replayed once it evicts from the dedupe cache. Unit-tested fresh/stale/skew/unparseable.

src/statichost/StaticHost/Live/LiveEndpoints.cs
IEvangelist: Fixed — the message-id dedupe now skips `webhook_callback_verification`, so a retried verification always re-echoes the challenge. Only real notifications dedupe.

src/statichost/StaticHost/Live/LiveEndpoints.cs
IEvangelist: Replaced the per-notification `Task.Run` with a `YouTubeLiveConfirmationQueue` background service: a bounded capacity-1 (DropWrite) channel coalesces a burst of retries into a single confirming poll, and it runs on the host's stopping token so it can't outlive shutdown or fan out concurrent polls.

src/statichost/StaticHost/Live/LiveStatusBroadcaster.cs
IEvangelist: Fixed — `Subscribe` now writes the seed and adds the writer under `_gate` in the same lock, so a flush can't slip between the two and leave the new subscriber stuck on a stale seed.

src/statichost/StaticHost/Live/LiveStatusBroadcaster.cs
IEvangelist: Yes — it now resolves the sticky primary against `basis.PrimarySource` (the pending snapshot), so during the coalescing window the source that arrived first stays primary instead of the second source winning against a stale `_current`. Added a regression test.

src/statichost/StaticHost/Live/LiveStatusBroadcaster.cs
IEvangelist: Fixed — the no-op check now compares with `UpdatedAt` excluded (`next with { UpdatedAt = _current.UpdatedAt } == _current`) in both `Update` and `FlushLocked`, so an identical reconcile no longer forces a timestamp-only broadcast. Added a test that a repeated identical update doesn't re-emit.

src/frontend/src/components/LiveVideosTabs.astro
IEvangelist: Fixed — the click listener now ignores untrusted events (`if (!event.isTrusted) return;`). The synthetic `activateTab()` click no longer sets `userSticky`, so automatic tab switching on primary-source changes keeps working; real user clicks (and Playwright's trusted clicks) still stick.

src/frontend/src/components/live-status.ts
IEvangelist: Fixed — the seed now only applies when its `updatedAt` isn't older than what's already rendered, so a slow `/api/live` response can't clobber a fresher snapshot the SSE stream already pushed.

src/frontend/tests/e2e/live-status.spec.ts
IEvangelist: You're right — the `ReadableStream` cast wasn't a real SSE body. Replaced all four with a shared `fulfillSseState` helper that fulfils the route with a complete string body (`event: state` + a single data line), so the real `EventSource` fires once deterministically. Tests that push further frames still use the mock `EventSource` + `__aspireLiveSseEmit` path.

src/frontend/scripts/build-static-host.mjs
IEvangelist: Fixed — the `scalar` and `.gitignore` restores now run in `finally`, before the temp dir is removed, so a failed `astro build` can't delete the only preserved copies.

src/statichost/StaticHost/Live/Twitch/TwitchEventSubService.cs
IEvangelist: Fixed — both the Twitch and YouTube workers now only swallow `OperationCanceledException` when `stoppingToken.IsCancellationRequested`; a transient cancellation (e.g. an HttpClient timeout) now falls through to the retry path instead of permanently stopping the loop.

src/apphost/Aspire.Dev.AppHost/LiveExtensions.cs
```diff
@@ -0,0 +1,357 @@
...
+    public static IResourceBuilder<ProjectResource> WithProductionLiveStatus(
+        this IResourceBuilder<ProjectResource> staticHostWebsite,
+        IDistributedApplicationBuilder builder)
+    {
+        var liveConfig = builder.AddAzureKeyVault("liveconfig");
+
+        var publicBaseUrl = liveConfig.AddSecret(
+            "live-public-base-url-secret",
+            "live-public-base-url",
+            builder.AddParameter("live-public-base-url", "https://aspire.dev", publishValueAsDefault: true));
```
eerhardt: Does this really need to be a secret? Seems more like a configuration setting.

src/apphost/Aspire.Dev.AppHost/LiveExtensions.cs
```diff
@@ -0,0 +1,357 @@
...
+            "live-twitch-channel-login",
+            builder.AddParameter("live-twitch-channel-login", "aspiredotdev", publishValueAsDefault: true));
+        var twitchChannelId = liveConfig.AddSecret(
+            "live-twitch-channel-id-secret",
+            "live-twitch-channel-id",
+            builder.AddParameter("live-twitch-channel-id"));
+        var twitchReconcileInterval = liveConfig.AddSecret(
+            "live-twitch-reconcile-interval-seconds-secret",
+            "live-twitch-reconcile-interval-seconds",
+            builder.AddParameter("live-twitch-reconcile-interval-seconds", "1800", publishValueAsDefault: true));
```
eerhardt: Does this need to be a secret?

src/apphost/Aspire.Dev.AppHost/LiveExtensions.cs
```diff
@@ -0,0 +1,293 @@
...
+            .WithEnvironment("Live__Twitch__ClientSecret", siteConfig.GetSecret("live-twitch-client-secret"))
+            .WithEnvironment("Live__Twitch__WebhookSecret", siteConfig.GetSecret("live-twitch-webhook-secret"))
+            .WithEnvironment("Live__Twitch__ChannelLogin", siteConfig.GetSecret("live-twitch-channel-login"))
+            .WithEnvironment("Live__Twitch__ChannelId", siteConfig.GetSecret("live-twitch-channel-id"))
+            .WithEnvironment("Live__Twitch__ReconcileIntervalSeconds", siteConfig.GetSecret("live-twitch-reconcile-interval-seconds"))
+            .WithEnvironment("Live__YouTube__ApiKey", siteConfig.GetSecret("live-youtube-api-key"))
+            .WithEnvironment("Live__YouTube__WebhookSecret", siteConfig.GetSecret("live-youtube-webhook-secret"))
+            .WithEnvironment("Live__YouTube__ChannelHandle", siteConfig.GetSecret("live-youtube-channel-handle"))
+            .WithEnvironment("Live__YouTube__ChannelId", siteConfig.GetSecret("live-youtube-channel-id"))
+            .WithEnvironment("Live__YouTube__PollingIntervalSeconds", siteConfig.GetSecret("live-youtube-polling-interval-seconds"))
```
eerhardt: Do these kinds of things need to be secrets? They seem more like configuration values.

eerhardt (approved):

src/apphost/Aspire.Dev.AppHost/LiveExtensions.cs
```diff
@@ -0,0 +1,332 @@
...
+            .WithEnvironment("Live__YouTube__DiscoveryPollingIntervalSeconds", youtubeDiscoveryPollingInterval)
+            .WithEnvironment("Live__YouTube__OfflineConfirmationCount", youtubeOfflineConfirmationCount)
+            .WithEnvironment("Live__Twitch__ClientSecret", siteSecrets.GetSecret("live-twitch-client-secret"))
+            .WithEnvironment("Live__Twitch__WebhookSecret", siteSecrets.GetSecret("live-twitch-webhook-secret"))
+            .WithEnvironment("Live__YouTube__ApiKey", siteSecrets.GetSecret("live-youtube-api-key"))
+            .WithEnvironment("Live__YouTube__WebhookSecret", siteSecrets.GetSecret("live-youtube-webhook-secret"))
+            .PublishAsAzureAppServiceWebsite((_, website) =>
+            {
+                // Live state and WebSub verification are coordinated in memory.
+                website.SiteConfig.NumberOfWorkers = 1;
```
eerhardt: What is this for? Does it screw our scaling up?

IEvangelist: Yes — good catch. That setting did cap the entire `aspiredev` web app to one App Service worker.

Aspire creates this App Service plan with `perSiteScaling = true` and normally gives each website a `numberOfWorkers` ceiling of 30 so it can use the plan's available instances. Setting it to 1 does not stop the plan itself from scaling or being billed for additional instances; it means this particular website can run on at most one of them. The current template does not configure plan autoscale, so it might not have changed the initial instance count, but it would have prevented future manual/autoscale expansion and multi-instance or zone resilience from benefiting the main site.

The reason it was present is that simply deleting it would make the current live implementation incorrect under scale-out. The live snapshot, `LiveSessionId`, coalescing timer, SSE subscriber channels, Twitch message replay cache, YouTube WebSub pending verification token and lease state, and YouTube confirmation queue are all process-local. With multiple workers:

- a webhook handled by worker A cannot notify SSE clients connected to worker B;
- a YouTube subscription initiated by A can have its verification callback routed to B, which has no matching pending token and returns 404;
- every worker runs its own Twitch/YouTube reconciliation and polling loops, causing duplicate subscription work, races, and additional provider quota usage; and
- clients can observe different snapshots or session IDs depending on which worker receives a request or reconnection.

I pushed `f0d742e4` to remove that tradeoff from the public site rather than just removing the guard. Production now deploys two websites in the same per-site-scaling App Service plan:

- `aspiredev` remains the Azure Front Door origin and keeps Aspire's normal 30-worker ceiling. It receives no live credentials, Key Vault access, broadcaster, or provider workers. It uses YARP to stream `/api/live/*` requests — including SSE, webhook bodies, headers, and query strings — to the coordinator.
- `aspiredev-live` owns the live configuration, read-only Key Vault references, provider workers, webhook state, and SSE broadcaster. Only this dedicated coordinator is limited to one worker for the current in-memory correctness requirement.

That isolates the limitation and failure domain: the documentation site remains horizontally scalable, while a coordinator outage affects only live-status requests. It also does not lock us into this topology. We can later move the snapshot and subscription state to shared storage, add cross-instance fan-out and a distributed lease/leader, then remove the coordinator's one-worker cap without changing the public `/api/live` contract or frontend.

IEvangelist: Follow-up: I completed the scale-out work in `777ebbac`, so the interim topology described above is now superseded.

There is no longer a single-worker live coordinator, a second `aspiredev-live` website, or YARP forwarding. Production is back to one `aspiredev` website behind Front Door, with Aspire's normal generated `numberOfWorkers: 30` ceiling and no custom worker restriction. The application architecture is now safe for any number of App Service workers within that capacity.

The shared coordination layer is Azure Managed Redis:

- **Canonical state:** Twitch and YouTube status are stored as one epoch-scoped, monotonically versioned snapshot. Updates run under a short distributed lock, and the final Redis write is fenced against the lock-owner token so a worker whose lease expired cannot overwrite newer state.
- **Cross-worker SSE:** each successful write publishes the complete snapshot through Redis pub/sub. Every worker keeps only a local SSE fan-out replica for its connected clients. New SSE connections reload canonical state first, and every worker resynchronizes every 30 seconds, so a missed publication or reconnect self-heals.
- **State-loss recovery:** versions are scoped to a random state epoch. If the canonical key is ever recreated, workers verify the new epoch against Redis before accepting it; delayed messages from the previous epoch cannot roll a worker back or leave SSE permanently stale.
- **Provider leadership:** Twitch reconciliation and YouTube polling/renewal each use an independent renewable Redis leadership lease. Only one worker performs each provider's background work at a time, but leadership automatically moves if that worker stops or loses its lease.
- **Twitch replay coordination:** message IDs are hashed and represented by distributed `processing` and `completed` states. A concurrent delivery receives `503` so Twitch retries; a completed replay receives `200`; failed processing releases its reservation.
- **YouTube WebSub coordination:** pending topic/token/request state, verification state, active topic, and renewal time are distributed. The subscription can be initiated on one worker and verified on any other worker. A short pre-send reservation allows fast takeover if the leader exits, and confirming polls are coalesced across workers.

Redis contains status and coordination data only. The Twitch client secret, Twitch webhook signing secret, YouTube API key, and YouTube webhook signing secret remain read-only Key Vault references. The Redis deployment uses Microsoft Entra authentication with access-key authentication disabled, and the website receives access through its managed identity.

Redis is intentionally required rather than having a silent in-memory production fallback, because a fallback would reintroduce split-brain state. A transient Redis failure no longer stops the StaticHost itself; existing SSE clients retain their last local snapshot, while operations requiring canonical state fail explicitly and recover when Redis returns.

I exercised this with two separately running StaticHost processes connected to the same Redis instance: an update accepted by one process reached an SSE client connected to the other, canonical state survived a process restart, repeated fenced writes remained readable through `IDistributedCache`, and deleting/recreating the canonical state key converged both workers onto the new epoch. The focused suites are also green: 138 StaticHost tests and 4 AppHost topology/configuration tests.

So the direct answer is now: **no, live status no longer screws up scaling**. We removed the one-worker workaround rather than relocating it; correctness, replay defense, subscriptions, provider ownership, and SSE fan-out are all coordinated across workers.

src/apphost/Aspire.Dev.AppHost/LiveExtensions.cs
IEvangelist: Addressed, including the internal deployment constraint you called out.

The non-sensitive values are ordinary deployment parameters. For the genuinely sensitive values, the AppHost now only grants **Key Vault Secrets User** and calls `WithReference(siteSecrets)`—there are no per-secret `AddSecret`/`GetSecret` resources or `Live__*` Key Vault environment expressions in the generated deployment.

StaticHost conditionally calls `builder.Configuration.AddAzureKeyVaultSecrets("siteconfig")` when the production vault connection is present. The vault secrets use the default configuration-provider naming convention:

- `Live--Twitch--ClientSecret`
- `Live--Twitch--WebhookSecret`
- `Live--YouTube--ApiKey`
- `Live--YouTube--WebhookSecret`

The `--` delimiters map to `:` and bind directly to the existing `LiveStatusOptions`. Local `aspire start` remains unchanged because it has no `siteconfig` connection and continues using the generated local-only webhook/dev-command secrets.

I also regenerated the deployment artifacts to confirm they contain the vault connection and managed-identity role assignment, but no `Microsoft.KeyVault/vaults/secrets` resources or `@Microsoft.KeyVault(...)` settings. This is in `80b27923`.

eerhardt (commented):
review the usage of key vault and redis

src/apphost/Aspire.Dev.AppHost/AppHost.cs
```diff
@@ -3,18 +3,33 @@
 // For deployment: We want to pick AppService as the environment to publish to.
 builder.AddAzureAppServiceEnvironment("production");
 
+var cache = builder.AddAzureManagedRedis("livecache")
```
eerhardt: it might make sense to just name this 'cache' in case we need to use it for something more in the future.

src/statichost/StaticHost/Program.cs
```diff
@@ -1,14 +1,26 @@
...
+builder.AddLiveStatus();
+
 builder.Services.AddHsts(options =>
 {
     options.Preload = true;
     options.IncludeSubDomains = true;
     options.MaxAge = TimeSpan.FromDays(180);
 });
 
+builder.Services.AddOpenApi("v1");
```
eerhardt: Should this only happen in development environments? Do we need OpenAPI in production?

How I guess we don't map the apis below in production.

src/statichost/StaticHost/Program.cs
```diff
@@ -1,14 +1,26 @@
+using Scalar.AspNetCore;
+
 var builder = WebApplication.CreateBuilder(args);
 
 builder.AddServiceDefaults();
 
+// Production supplies the vault reference; local Aspire runs intentionally do not.
+if (builder.Configuration.GetConnectionString("siteconfig") is not null)
```
eerhardt: (nit) I don't think `siteconfig` makes much sense for a key vault name. How about "secrets"? or "kv" / "keyVault"?

src/statichost/StaticHost/Program.cs
```diff
@@ -30,7 +42,16 @@
 // LinkHeaderMiddleware attaches a Link header on HTML 2xx responses via OnStarting.
 app.UseAgentReadiness();
 
+var staticAssetsManifest = Path.Combine(
+    AppContext.BaseDirectory,
+    $"{app.Environment.ApplicationName}.staticwebassets.endpoints.json");
+var hasStaticAssetsManifest = File.Exists(staticAssetsManifest);
+
 app.UseDefaultFiles();
+if (!hasStaticAssetsManifest)
```
eerhardt: When do we expect this to not be there?

src/statichost/StaticHost/Live/LiveStatusStore.cs
```diff
@@ -0,0 +1,215 @@
...
+
+            _state = new LiveStatusState(_state.Epoch, _state.Version + 1, next);
+            return ValueTask.FromResult(_state);
+        }
+    }
+}
+
+internal sealed class RedisLiveStatusStore(
+    IDistributedCache cache,
+    IConnectionMultiplexer connectionMultiplexer,
```
eerhardt: It's kind of weird to take both an IDistributedCache and a IConnectionMultiplexer. I would think we could just lean in on Redis - we don't really need the IDistributedCache abstraction IMO.

src/statichost/StaticHost/Live/RedisDistributedLock.cs
```diff
@@ -0,0 +1,85 @@
...
+
+namespace StaticHost.Live;
+
+internal sealed class RedisDistributedLock(
+    IConnectionMultiplexer connectionMultiplexer,
+    TimeProvider timeProvider,
+    ILogger<RedisDistributedLock> logger)
+{
+    private static readonly TimeSpan s_lockLifetime = TimeSpan.FromSeconds(15);
+    private static readonly TimeSpan s_acquireTimeout = TimeSpan.FromSeconds(5);
```
eerhardt: Could we use optimistic concurrency with bounded retries for the live-status and YouTube subscription records instead of this lease-based lock? The protected operations are short read–modify–write updates, with no external API calls inside the lock. `LiveStatusState` already has an epoch/version pair: atomically write only if both still match, then reread and recompute on conflict. The subscription record could use a similar version envelope or compare the previously read value.

This would remove the lock acquisition/release and lease-expiry handling. Using application-owned Redis records would also avoid writing `IDistributedCache`'s internal hash fields. We'd need conditional creation for missing keys, bounded/cancellable retries, and publication only after a successful write. The separate provider-leadership leases and webhook deduplication reservations should remain; they coordinate work rather than state updates.

src/statichost/StaticHost/wwwroot/scalar/aspire-theme.css
```diff
@@ -0,0 +1,67 @@
...
+ * Custom Scalar theme for aspire.dev — aligned to the Aspire brand kit
+ * (https://aka.ms/aspire/brand). Mirrors the tokens used by the main site
+ * in src/frontend/src/styles/site.css so the API reference looks at home
+ * inside the dashboard.
+ */
+
+:root {
+  --aspire-color-primary: #7455dd;
+  --aspire-color-light: #dcd5f6;
+  --aspire-color-dark: #1f1e33;
```
eerhardt: Could we extract the common brand colors and font-family tokens into a shared stylesheet used by both the frontend and Scalar, keeping only the Scalar-specific mappings and selectors here? These colors already exist in `src/frontend/src/styles/site.css`, and the typography is already diverging: this theme prefers Outfit while the site uses Poppins. A single source for the shared tokens, wired into both the frontend build and StaticHost's development-time CSS loading, would avoid maintaining two copies without pulling all of the site's CSS into Scalar.

src/statichost/StaticHost/Live/LiveStatusBroadcaster.cs
```diff
@@ -0,0 +1,256 @@
...
+/// </remarks>
+/// <remarks>Creates the broadcaster.</remarks>
+public sealed class LiveStatusBroadcaster(
+    IOptions<LiveStatusOptions> options,
+    ILogger<LiveStatusBroadcaster> logger,
+    TimeProvider? timeProvider = null,
+    ILiveStatusStore? store = null) : IDisposable
+{
+    private readonly TimeProvider _time = timeProvider ?? TimeProvider.System;
+    private readonly ILiveStatusStore _store = store ?? new InMemoryLiveStatusStore(timeProvider);
```
eerhardt: Do we ever use the InMemoryLiveStatusStore? we always register a ILiveStatusStore service, so this shouldn't ever be used, right?

src/apphost/Aspire.Dev.AppHost/AppHost.cs
IEvangelist: Done — I renamed the actual Aspire resource from `livecache` to `cache`, not just the local variable. StaticHost now resolves `AddRedisClientBuilder("cache")`, AppHost tests assert the resource and `ConnectionStrings__cache`, and the generated deployment uses `cache/cache.bicep` plus `aspiredev-roles-cache`. That keeps the resource name intentionally general so the same Redis service can support additional caching or coordination uses later.

src/statichost/StaticHost/Program.cs
IEvangelist: Yes — registration now matches endpoint exposure. `AddOpenApi("v1")` is guarded by `builder.Environment.IsDevelopment()`, while `MapOpenApi()` and Scalar remain under the existing development-only application guard. Production therefore neither registers the OpenAPI document services nor maps the document/UI endpoints.

src/statichost/StaticHost/Program.cs
IEvangelist: Agreed and fixed. The Key Vault resource and connection are now named `secrets`; production loads it with `AddAzureKeyVaultSecrets("secrets")`, and the deployment emits `ConnectionStrings__secrets` and `aspiredev-roles-secrets`. The vault still contains only the four sensitive provider/API/signing values, while the eleven ordinary `Live__*` settings remain non-secret deployment parameters.

src/statichost/StaticHost/Program.cs
IEvangelist: In a supported Web SDK build, we do not expect the manifest to be absent. That branch was compensating for the PR-local Debug `<Content Remove="wwwroot\**" />`, which removed generated frontend files from the static-web-asset graph. I removed both that ItemGroup and the manifest probe/fallback. StaticHost now unconditionally calls `MapStaticAssets()` like `main`, avoiding a second physical-file serving path that could bypass the endpoint cache metadata in `Caching.targets`.

src/statichost/StaticHost/Live/LiveStatusStore.cs
IEvangelist: Done — the live-status implementation now leans directly on Redis. I removed `IDistributedCache`, `.WithDistributedCache()`, and the `Aspire.StackExchange.Redis.DistributedCaching` package. The application owns plain JSON Redis string records through a small direct Redis adapter used for `GET`, atomic compare-and-set, and pub/sub. I also moved the record keys to `v2` so an existing `v1` cache-format hash cannot cause `WRONGTYPE` during rollout; a runtime check confirmed the canonical state is now a Redis `string`.

src/statichost/StaticHost/Live/RedisDistributedLock.cs
IEvangelist: Implemented this as bounded optimistic concurrency and removed `RedisDistributedLock` entirely. Each short record update now runs a single-key Lua compare-and-set against the exact bytes read: missing-key creation is conditional, a contending writer receives the winner's current value, re-applies its provider-specific transition, and retries up to eight times with capped exponential jitter and cancellation. The live-state epoch is created canonically on first touch, versions advance only after a successful CAS, and pub/sub happens only after the write commits. YouTube subscription state uses a versioned envelope as an additional ABA guard, and a request returned by `TryBeginSubscriptionAsync` is always the request from the winning attempt. On exhaustion the webhook paths return `503` so the provider retries. I retained the renewable provider-leadership leases, Twitch processing/replay reservations, and YouTube confirmation coalescing key because those are genuine distributed coordination lifetimes rather than short record updates. Added focused conflict/no-op/cancellation/version tests, and an eight-way real Redis contention run completed 8/8 requests successfully with both provider updates merged.

src/statichost/StaticHost/wwwroot/scalar/aspire-theme.css
IEvangelist: Done. `src/frontend/src/styles/aspire-brand.css` is now the single source for shared Aspire colors, action tokens, shadows, and the Poppins/Fira Code font stacks. `site.css` imports those tokens, while `aspire-theme.css` contains only Scalar mappings, dark-mode overrides, and the live-tag selector — no duplicated hex colors. StaticHost copies the shared token file into its output and prepends it when supplying Scalar's custom CSS. Scalar now uses `var(--aspire-font-sans)` with Poppins first, and I removed the unused Outfit webfont from Starlight configuration, `package.json`, and the lockfile. Frontend lint/component tests pass, and the live Scalar document contains the shared tokens with no Outfit reference.

src/statichost/StaticHost/Live/LiveStatusBroadcaster.cs
IEvangelist: Correct — production DI always supplies the store, so the fallback was both unused and dangerous. `LiveStatusBroadcaster` now requires non-null `TimeProvider` and `ILiveStatusStore`; no production in-memory store remains. I applied the same fail-fast rule to provider coordination and YouTube subscription state, since silently falling back there would also disable cross-instance leadership/deduplication. The in-memory implementations now live only in `StaticHost.Tests`, and a DI registration test asserts that production registers `LiveStatusRedis`/`RedisLiveStatusStore` and no `IDistributedCache`.

mitchdenny:
## Review findings

Reviewed commit `0e3a7f540bb84987a32bf280f9467ef5bb298318`. These issues should be addressed before merging.

### High

1. **Client-side navigation removes the unanswered cookie-consent banner.**  
   [Head.astro:43](https://github.com/microsoft/aspire.dev/blob/0e3a7f540bb84987a32bf280f9467ef5bb298318/src/frontend/src/components/starlight/Head.astro#L43): the new global `ClientRouter` replaces the dynamically inserted WCP banner during body swaps, while `window.__aspireWcpConsentInitialized` survives and prevents initialization from restoring it. Visitors in consent-required regions who navigate before answering lose the banner. Preserve a persistent consent host and its required root state across navigation, and add a navigate-before-consenting regression test. This was reproduced with the actual consent initializer in a simulated body swap.

2. **A stale YouTube search can overwrite a newer live observation.**  
   [LiveEndpoints.cs:417-420](https://github.com/microsoft/aspire.dev/blob/0e3a7f540bb84987a32bf280f9467ef5bb298318/src/statichost/StaticHost/Live/LiveEndpoints.cs#L417-L420): webhook confirmation unconditionally publishes its search result. An offline search can start, the polling worker can discover and publish a newly started broadcast, and the older empty response can then overwrite it with `(false, null)`. The worker no longer performs known-video checks and can wait until its next discovery deadline, up to 30 minutes away. The positive polling write has the corresponding stale-result race. Apply results conditionally against the provider state observed before the request, share offline-confirmation logic, and test these request interleavings.

3. **A lost YouTube verification response can leave the subscription inactive for days.**  
   [YouTubeWebSubSubscriptionState.cs:131-144](https://github.com/microsoft/aspire.dev/blob/0e3a7f540bb84987a32bf280f9467ef5bb298318/src/statichost/StaticHost/Live/YouTube/YouTubeWebSubSubscriptionState.cs#L131-L144): confirmation clears `Pending` and schedules renewal before the challenge response reaches the hub. If that response is lost, matching verification retries return 404 because the pending token is gone. The hub cannot activate the subscription, but the application defers another request until renewal, four days later for a five-day lease. Retain recently confirmed topic/token data and accept bounded idempotent retries without repeatedly extending the lease. Cover a retry on another worker after response loss.

4. **Prematurely invalidated Twitch tokens remain cached.**  
   [TwitchAppTokenProvider.cs:26-40](https://github.com/microsoft/aspire.dev/blob/0e3a7f540bb84987a32bf280f9467ef5bb298318/src/statichost/StaticHost/Live/Twitch/TwitchAppTokenProvider.cs#L26-L40): a token rejected before its advertised expiry continues to be returned on every reconciliation. Helix 401 responses never invalidate or reacquire it, so polling and subscription repair keep failing until expiry or process restart. On 401, conditionally invalidate the rejected token, obtain a fresh app token, and retry once. Add an early-invalidation regression test.

### Medium

5. **C# and TypeScript API search controllers interfere after navigation.**  
   [TypeScript module page:662](https://github.com/microsoft/aspire.dev/blob/0e3a7f540bb84987a32bf280f9467ef5bb298318/src/frontend/src/pages/reference/api/typescript/%5Bmodule%5D/index.astro#L662): both languages retain `astro:page-load` controllers targeting the same `pkg-*` elements and shared globals. After C# package -> TypeScript module -> C# package navigation, the TypeScript renderer replaces C# search links with non-clickable elements because C# results lack its expected `h` URL field. Guard initialization with language-specific page/root ownership, including the type/member controllers. This was reproduced by executing the actual controller scripts; add a cross-language navigation regression test.

6. **The backend test suites are not executed by CI.**  
   [Aspire.Dev.slnx:9](https://github.com/microsoft/aspire.dev/blob/0e3a7f540bb84987a32bf280f9467ef5bb298318/Aspire.Dev.slnx#L9): adding the AppHost test project to the solution does not make the current workflows run it. `apphost-build.yml` only builds the AppHost; `tools-tests.yml` only runs the two generator suites. Wire `Aspire.Dev.AppHost.Tests` and `StaticHost.Tests` into CI and include their test paths in change detection.

### Test coverage

Local runs passed: 33 focused live-infrastructure tests and four AppHost tests after restore/build, plus 46 provider tests using the resulting binaries with `--no-build --no-restore`. Existing CI checks pass, including desktop/tablet/mobile e2e. The scenarios above lack regression coverage. Mapped live HTTP endpoints and production Redis coordination/synchronizer behavior are not covered by the current tests; coordination tests exercise the in-memory fake. Axe audits omit the videos page and open live picker.

The frontend reproductions were dependency-free simulations, not browser runs. Frontend dependencies were absent, so local Vitest, browser tests, and typechecking were not run. No local frontend build was run.

IEvangelist:
@mitchdenny Addressed the agreed feedback in cc7b01021dd8d03b80a76015e0874e2057709c37.

- Preserve the original WCP consent host, runtime, vendor styles, root consent state, and banner offset across client navigation, with navigate-before-consenting coverage.
- Fence YouTube results atomically using provider-specific epoch/revision metadata in Redis. Polling and webhook checks now share offline evidence; stale positive/negative and overlapping same-state results cannot overwrite newer observations.
- Persist recently confirmed WebSub tokens for idempotent cross-worker retries, bounded by the original lease or ten minutes. Retries never extend renewal, and replacement subscriptions invalidate old tokens.
- Conditionally invalidate rejected Twitch tokens and retry every Helix operation once with a fresh token, without discarding concurrent replacements.
- Give all six API search surfaces language/page-owned roots, local index data, and abortable controller/timer lifecycles.
- Run both backend suites and mandatory real-Redis integration coverage in the existing AppHost CI job, with the test directories and solution included in change detection.

Also added mapped HTTP endpoint tests and live videos/open-picker axe and keyboard coverage. Redis tests use an explicit disposable instance and exact-key cleanup, never a shared/default database or database-wide flush.

Local results: 236 StaticHost tests (including 18 against Redis 8.6), 4 AppHost tests, and 107 focused frontend unit tests passed. Consent/live/accessibility coverage passed 62 browser cases across desktop/tablet/mobile, with 7 existing device-specific skips. All 12 API navigation cases passed using native transitions on desktop and the supported ClientRouter swap fallback on touch projects. Native touch emulation surfaced Chromium `InvalidStateError` transition-capture aborts; those were not hidden by filtering page errors, and the explicit fallback cases still assert preserved JavaScript context and clickable correct-language results.

The repository-wide TypeScript check has existing failures: baseline 187 diagnostics, current 185, with zero newly introduced diagnostics. No local frontend production build was run; the updated commit has triggered CI.

.github/workflows/apphost-build.yml
```diff
@@ -13,6 +13,17 @@ jobs:
     env:
       DOTNET_NOLOGO: 1
       DOTNET_CLI_TELEMETRY_OPTOUT: 1
+      TEST_RESULTS_DIR: ${{ github.workspace }}/testresults
+    services:
+      redis:
+        image: redis:8.6
```
eerhardt: Why can't we use aspire for the tests? That way this image stays up to date.

IEvangelist: Done

PR #859, opened 2026-05-06
fix: update keyboard shortcut for stopping AppHost in documentation

Jeffreyyvdb:
Stopping the apphost is done with  ⌃ (control) + c on mac instead of ⌘ + c


Jeffreyyvdb:
@microsoft-github-policy-service agree

davidfowl (approved):

PR #884, opened 2026-05-07
Style inline code with theme-aware accent + hairline border

IEvangelist:
## Summary

Restyles inline `<code>` so it reads as a polished accent chip instead of a flat grey block, and makes the chip **adapt to its surrounding surface**:

- **Hairline border** using `currentColor` (auto-tracks the chip's text color)
- **Theme-aware accent fill** built from `currentColor` mixed with a neutral
  - Dark: `color-mix(in srgb, currentColor 22%, #000)` &rarr; darker shade of the accent
  - Light: `color-mix(in srgb, currentColor 11%, #fff)` &rarr; lighter tint of the accent
- **On-theme text color** drawn from Starlight's accent variables
- **Adopts the surrounding accent** automatically: a chip inside a note aside reads blue, inside a tip aside it reads green, inside caution it reads orange, inside danger it reads red &mdash; everywhere else it uses the page's mauve/purple accent

The padding and pill `border-radius` already exist in the sibling Starlight rules, so this PR only adds what was missing (color, bg, border).

## Why `currentColor + neutral` instead of `accent + page bg`

The original mix `color-mix(accent, var(--sl-color-bg))` resolved the chip surface against the **page** background. In Catppuccin Mocha that base is `#1f1e33` (purple-toned), so a chip dropped inside an orange caution aside still came out purple-grey &mdash; visibly wrong against the orange surrounding surface. Switching the formula to `currentColor + #000`/`#fff` makes the chip a darker/lighter shade of its **own text color**, so it always blends naturally with whatever accent it's inheriting (page or aside).

A nice side-effect: per-variant overrides only need to set `--aspire-inline-code-text`. Bg and border follow automatically via `currentColor`.

## Special case: aside link pills

Starlight gives `<a>` inside an aside its own filled background and pill `border-radius` (so links read as buttons). When a code chip sat inside that link (`<a><code>...</code></a>`), the link's pill stacked **under** the new chip producing a doubled border / wider light surface. A targeted rule clears the link pill **only** when its sole child is `<code>`, leaving normal link styling untouched.

## WCAG 2 AA verification

Computed contrast ratios &mdash; chip text against the chip's resolved background:

| Surface     | Dark mode | Light mode |
|-------------|-----------|------------|
| Page (mauve / brand) | **7.32** | **10.80** |
| Note aside (blue)    | **7.12** | **7.21**¹ |
| Tip aside (green)    | **9.10** | **4.69**  |
| Caution aside (orange) | **8.12** | **5.12**² |
| Danger aside (red)   | **6.63** | **4.52**  |

All values are above the 4.5 AA threshold for body text; most also clear the 7.0 AAA bar.

¹ Note in light mode pins to `#1e40af` (darker than Catppuccin Latte's default note blue, which dipped below AA on a self-tinted surface).
² Caution in light mode pins to `#ab3f00` (already used elsewhere in the site for the caution title).

## Screenshots

> Hosted at https://github.com/IEvangelist/aspire.dev-1/releases/tag/pr-884-screenshots

### Before

| Dark | Light |
|------|-------|
| before dark | before light |

### After &mdash; page-level inline code

The body chips (`aspire --version`, `AddProject<TProject>(...)`, etc.) now wear the brand accent.

page level dark

### After &mdash; alert-aware accent

Each aside variant restyles inline code in its own accent. Note=blue, tip=green, caution=orange, danger=red.

| Dark | Light |
|------|-------|
| after asides dark | after asides light |

### Real-world example: Troubleshooting aside on the *Install Aspire CLI* page

The `aspire --version` chip cleanly adopts the caution accent in both themes &mdash; no more purple chip inside an orange surface.

| Dark | Light |
|------|-------|
| troubleshooting dark | troubleshooting light |

### DIFF
--- a/src/frontend/src/styles/site.css
+++ b/src/frontend/src/styles/site.css
@@ -291,6 +291,20 @@
   color: inherit;
 }
 
+/* Aside links normally render as a tinted pill (padding + bg + radius),
+   but when the link's sole child is inline code the chip already carries
+   that styling — stacking both produces a wider link pill with a second,
+   lighter border around the code chip. Suppress the link pill in that
+   case so only the code chip is visible.
+   :is(:root[data-theme='light']) anchors specificity high enough to
+   beat both the dark (:root …) and light (:root[data-theme='light'] …)
+   variants of the aside-link rule. */
+:is(:root[data-theme='light'], :root) .starlight-aside__content a:not([role='tab']):has(> code:only-child) {
+  background: none;
+  padding: 0;
+  border-radius: 0;
+}
+
 .poppins-light {
   font-family: 'Poppins', sans-serif;
   font-weight: 300;
--- a/src/frontend/src/styles/site.css
+++ b/src/frontend/src/styles/site.css
@@ -49,21 +49,20 @@
   --color-yellow: hsl(50, 55%, 75%);    /* catppuccin-aligned pastel gold */
 
   /* Inline code surface (dark theme).
-     bg  = #1f1e33 + 14% accent  ≈ #2b264b   (luma ≈ 0.024)
-     fg  = --sl-color-accent-high (#c6c2f2)  → contrast ≈ 8.4:1 (AAA)
-     The border alpha keeps the hairline visible without competing with text. */
-  --aspire-inline-code-bg: color-mix(in srgb, var(--sl-color-accent) 14%, var(--sl-color-bg));
+     bg uses currentColor (= the chip's text accent) mixed with #000 so the
+     chip surface always reads as a darker shade of the accent, regardless
+     of what surface it sits on (page bg, aside bg, etc.). Borders use
+     currentColor-on-transparent so they overlay whatever's behind them.
+     The 22% mix keeps text contrast > 6:1 (AAA) for every accent we use. */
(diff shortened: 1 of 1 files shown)


mitchdenny (approved):
Great work on this! The `currentColor`-based approach is really elegant — the bg and border automatically follow the text accent without needing extra selectors, and the comments throughout do an excellent job explaining the *why*.

A few things I noticed (caveat: I'm no CSS guru, so take these with a grain of salt):

- The hard-coded accessibility pins (`#1e40af`, `#ab3f00`) for light-theme aside variants could silently break if Starlight or Catppuccin Latte update their base colors. Might be worth documenting the Starlight version they target or extracting them as named vars.
- Nesting `.starlight-aside` rules inside `:root` is valid CSS Nesting but a little unconventional — a short comment clarifying this is intentional would help future maintainers.
- The specificity hack (`:is(:root[data-theme='light'], :root) .starlight-aside__content a...`) is well-commented, but a note like `/* revisit on Starlight upgrade */` would be a nice safety net.

None of these are blockers from where I'm sitting. The implementation is solid and well thought out. LGTM! 🚀

PR #891, opened 2026-05-08
Improve site search UX with keyboard navigation, footer hints, and query-forwarding API buttons

IEvangelist:
Improves the Pagefind-backed site search dialog with keyboard navigation, a hint footer, and query-forwarding API buttons.

## Screenshots

| Empty (just opened) | After typing |
|---|---|
| Empty dialog - only the C#/TypeScript API buttons are present and nothing is auto-highlighted | After typing browser - the first result auto-highlights with a single accent outline; sub-articles below stay quiet |

| Arrow-key navigation | Arrow nav reaches the API buttons |
|---|---|
| After ArrowDown ArrowDown - active highlight has moved to a sub-article with the same single-outline treatment, and scroll keeps it clear of the sticky keyboard footer | Pressing End jumps to the TypeScript API button at the very end of the nav order - same accent treatment as the result rows |

## What changed

src/frontend/src/components/starlight/Search.astro

- **Keyboard navigation**: ArrowDown / ArrowUp / Home / End / Enter walk through Pagefind results. Up/Down also extend through the **Load more results** button and the **C# / TypeScript API buttons** so arrow nav covers the same set as native Tab. Escape closes the dialog (native `<dialog>` behaviour).
- **Unified Tab + arrow nav**: keydown lives on the dialog at capture phase so arrows fire whether focus is on the input or on a focused result. Arrows move *real* focus (`preventScroll: true`); a `focusin` handler keeps the active index in sync after Tab so up/down resume cleanly from wherever Tab landed.
- **Auto-highlight is restricted to result links**: opening the dialog with no query leaves nothing highlighted; as soon as Pagefind renders, the first result auto-highlights (DocSearch / GitHub command-palette style). Buttons are reachable via arrows but never default-active, so the first ArrowDown after typing always lands on the next result.
- **Active marker matches Starlight's existing `:focus-within`** exactly (1px accent-high outline + accent-low background), so virtual highlight (input still focused) and real focus produce identical pixels - no double border, no competing rules.
- **Scroll follows nav**: `scroll-margin-block-end` clears the sticky keyboard footer so `scrollIntoView({block: 'nearest'})` keeps the active row in view as you arrow through long result lists.
- **Pagefind re-renders coalesced via `requestAnimationFrame`** so the active highlight no longer flickers as the user types.
- **A11y wiring**: input gets `role=combobox`, `aria-controls`, `aria-expanded` (now reflects `dialog.open` - flips to `false` on Esc / backdrop close), and `aria-activedescendant` (only points at result-link options, not buttons). Result links get `role=option` + `aria-selected`. Results container gets `role=listbox` and an id that is force-set on every refresh so `aria-controls` is never broken by an upstream id assignment.
- **Footer keyboard hints**: sticky footer at the bottom of the dialog with arrow keys for navigation, Enter to open, Esc to close. Responsive layout collapses gracefully on mobile; `prefers-reduced-motion` is honoured.
- **API button query forwarding**: the C# and TypeScript API anchors in `api-search-notice-tpl` propagate the typed query as `?q=<encoded>`. `inpage-search-sync.ts` already consumes that on the landing pages, so this is end-to-end with no consumer changes. Default `href` is restored when the input is cleared, so middle/right-click `open in new tab` and no-JS scenarios keep working.
- **Single `MutationObserver`** drives template injection (idempotent) and refreshes both the keyboard-nav target list and the API-button `href`s whenever Pagefind re-renders. Re-initialised on `astro:after-swap`.

## Tests

`src/frontend/tests/e2e/site-search.spec.ts`

- Footer hints render with all `<kbd>` shortcuts.
- **Arrow keys cycle through results, Tab keeps state, and Escape closes the dialog**: ArrowDown moves real focus to the next result, Tab advances both focus and the active marker, ArrowUp resumes from the post-Tab position, and Escape both hides the dialog and flips `aria-expanded` back to `false`.
- **Arrow nav also cycles through the C#/TypeScript API buttons**: `End` jumps to the TS button (last in nav order), ArrowUp lands on the C# button, `Home` returns to the first result and clears the marker on the buttons.
- Query forwarding: typing `withBun` rewrites both API anchor `href`s, and clicking the TS link lands on `/reference/api/typescript/?q=withBun` with `#ts-api-search-input` populated and `.api-search-result` visible.
- Clearing the input restores default `href`s.

The `openSearchDialog()` helper `test.skip()`s when Pagefind isn't mounted (i.e., in `astro dev`), since the index is only built for `astro preview` / production. CI runs against the built site, so it exercises the full path.

## Validation

- `pnpm lint` - clean (`--max-warnings 0`)
- `pnpm test:unit` - all suites pass
- `pnpm build:production` - completes (Pagefind index rebuilt over 12,939 HTML pages)
- Local `astro preview` smoke-tested through every feedback round: empty-state, type-to-search, arrow nav across results + buttons, Tab/arrow sync, Escape, scroll-follow, query forwarding.

## Notes

- Pagefind itself is unchanged; everything is layered on top via DOM + a small `MutationObserver`, so this is robust to upstream Starlight / Pagefind updates.
- `:has()` is used for the active-result highlight selector (supported in all evergreens since 2023, matches the rest of the site).
- Auto-highlighting the first result follows Algolia DocSearch and the GitHub command palette. If we'd rather require `ArrowDown` first, that's a one-line change.

### DIFF
--- a/src/frontend/src/components/starlight/Search.astro
+++ b/src/frontend/src/components/starlight/Search.astro
@@ -120,6 +120,10 @@ const typescriptHref = `${base}/reference/api/typescript/`;
         input.setAttribute('role', 'combobox');
         input.setAttribute('aria-autocomplete', 'list');
         input.setAttribute('aria-controls', this.resultsContainerId);
+        // Seed `aria-expanded` so it's always present once the combobox
+        // role is set; the line below this `if` will refresh it to the
+        // real dialog state on every render.
+        input.setAttribute('aria-expanded', 'false');
         // Capture-phase keydown on the dialog catches arrows whether the
         // user is in the input or already on a focused result/button.
         this.dialog.addEventListener('keydown', this.keydownHandler, true);
@@ -168,9 +172,23 @@ const typescriptHref = `${base}/reference/api/typescript/`;
           ? (this.targets[this.activeIndex] as HTMLAnchorElement)
           : null;
 
+      // Clean up only the elements that are leaving the target list.
+      // Elements that survive into `newTargets` keep their attributes so
+      // a transient Pagefind re-render (which can briefly emit a target
+      // list without the result links) doesn't permanently wipe the
+      // active highlight — `setActive` below will re-establish state on
+      // the surviving element if appropriate.
+      const newSet = new Set(newTargets);
       for (const stale of this.targets) {
+        if (newSet.has(stale)) continue;
         stale.removeAttribute('data-search-active');
-        stale.setAttribute('aria-selected', 'false');
+        // `aria-selected` is only valid on roles in the listbox family
+        // (option/row/tab/gridcell/treeitem). The load-more / API buttons
+        // sit outside the listbox, so we never write the attribute on
+        // them — and we don't clear it on them either.
+        if (stale.getAttribute('role') === 'option') {
+          stale.setAttribute('aria-selected', 'false');
+        }
       }
       this.targets = newTargets;
 
@@ -222,7 +240,9 @@ const typescriptHref = `${base}/reference/api/typescript/`;
       if (this.activeIndex >= 0 && this.targets[this.activeIndex]) {
         const prev = this.targets[this.activeIndex];
         prev.removeAttribute('data-search-active');
-        prev.setAttribute('aria-selected', 'false');
+        if (prev.getAttribute('role') === 'option') {
+          prev.setAttribute('aria-selected', 'false');
+        }
       }
       this.activeIndex = index;
       if (index < 0 || !this.targets[index]) {
@@ -231,10 +251,14 @@ const typescriptHref = `${base}/reference/api/typescript/`;
       }
       const el = this.targets[index];
       el.setAttribute('data-search-active', 'true');
-      el.setAttribute('aria-selected', 'true');
-      // Only point aria-activedescendant at result links; the API/load-more
-      // buttons aren't part of the listbox.
+      // `aria-selected` and `aria-activedescendant` are only meaningful
+      // when the target lives inside the listbox (role="option"). The
+      // load-more button and the API reference buttons are plain
+      // focusable controls, so we deliberately leave both attributes
+      // off them — otherwise screen readers may announce them as
+      // "selected" listbox options.
       if (el.getAttribute('role') === 'option') {
+        el.setAttribute('aria-selected', 'true');
         this.input?.setAttribute('aria-activedescendant', el.id);
       } else {
         this.input?.removeAttribute('aria-activedescendant');
@@ -261,10 +285,14 @@ const typescriptHref = `${base}/reference/api/typescript/`;
         e.preventDefault();
         const prev = this.activeIndex <= 0 ? this.targets.length - 1 : this.activeIndex - 1;
         this.setActive(prev, true);
-      } else if (e.key === 'Home' && this.targets.length > 0) {
+      } else if (e.key === 'Home' && this.targets.length > 0 && e.target !== this.input) {
+        // Home/End are intercepted only when focus has already moved out
+        // of the input. While the caret is in the search field, the
+        // browser's native cursor behaviour (jump to start/end of the
+        // query text) takes priority over list navigation.
         e.preventDefault();
         this.setActive(0, true);
-      } else if (e.key === 'End' && this.targets.length > 0) {
+      } else if (e.key === 'End' && this.targets.length > 0 && e.target !== this.input) {
         e.preventDefault();
         this.setActive(this.targets.length - 1, true);
       } else if (e.key === 'Enter' && this.activeIndex >= 0 && this.targets[this.activeIndex]) {
@@ -515,6 +543,15 @@ const typescriptHref = `${base}/reference/api/typescript/`;
   }
 
   /* ── Keyboard hints footer ─────────────────────────────────────── */
+  /*
+   * The footer is rendered inside Starlight's `.dialog-frame`, which
+   * applies `padding: 1rem` (and `1.5rem` at >= 50rem). The negative
+   * `margin-inline` / `margin-block-end` values and the matching sticky
+   * `bottom` offset below "bleed" the footer back to the dialog edges so
+   * the border-top runs full width. If Starlight changes those padding
+   * tokens, update these values to match — otherwise the footer will
+   * appear visually inset from the dialog frame.
+   */
   .search-keyboard-footer {
     display: flex;
     flex-wrap: wrap;
--- a/src/frontend/src/components/starlight/Search.astro
+++ b/src/frontend/src/components/starlight/Search.astro
@@ -9,6 +9,8 @@ const pageId = routeData?.id ?? '';
 const isApiRefPage = pageId.startsWith('reference/api') || Astro.url.pathname.startsWith('/reference/api');
 const hideSearch = isApiRefPage;
 const base = import.meta.env.BASE_URL.replace(/\/$/, '');
+const csharpHref = `${base}/reference/api/csharp/`;
+const typescriptHref = `${base}/reference/api/typescript/`;
 ---
 
 { hideSearch ? <></> : <DefaultSearch><slot /></DefaultSearch> }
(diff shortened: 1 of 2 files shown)


src/frontend/src/components/starlight/Search.astro
```diff
@@ -17,52 +19,392 @@ const base = import.meta.env.BASE_URL.replace(/\/$/, '');
...
+        this.setActive(next, /* moveFocus */ true);
+      } else if (e.key === 'ArrowUp') {
+        if (this.targets.length === 0) return;
+        e.preventDefault();
+        const prev = this.activeIndex <= 0 ? this.targets.length - 1 : this.activeIndex - 1;
+        this.setActive(prev, true);
+      } else if (e.key === 'Home' && this.targets.length > 0) {
+        e.preventDefault();
+        this.setActive(0, true);
+      } else if (e.key === 'End' && this.targets.length > 0) {
```
JamesNK: **Bug:** Home/End keys are intercepted at capture phase on the dialog, so pressing Home/End while typing in the search input steals focus away instead of moving the cursor within the query text. This breaks standard text editing behavior.

```suggestion
      } else if (e.key === 'Home' && this.targets.length > 0 && e.target !== this.input) {
        e.preventDefault();
        this.setActive(0, true);
      } else if (e.key === 'End' && this.targets.length > 0 && e.target !== this.input) {
```

src/frontend/src/components/starlight/Search.astro
```diff
@@ -17,52 +19,392 @@ const base = import.meta.env.BASE_URL.replace(/\/$/, '');
...
+        prev.setAttribute('aria-selected', 'false');
+      }
+      this.activeIndex = index;
+      if (index < 0 || !this.targets[index]) {
+        this.input?.removeAttribute('aria-activedescendant');
+        return;
+      }
+      const el = this.targets[index];
+      el.setAttribute('data-search-active', 'true');
+      el.setAttribute('aria-selected', 'true');
```
JamesNK: **A11y:** `aria-selected` is set on all navigable targets, but it's only valid on elements with `role="option"`, `row`, `tab`, `gridcell`, or `treeitem`. The "Load more results" button and API buttons don't have any of those roles, so `aria-selected` on them may cause confusing screen reader announcements.

Consider guarding with:
```ts
if (el.getAttribute('role') === 'option') {
    el.setAttribute('aria-selected', 'true');
}
```

Same guard should apply in the cleanup loop above (line ~178) that sets `aria-selected` to `'false'` on stale targets.

src/frontend/src/components/starlight/Search.astro
```diff
@@ -17,52 +19,392 @@ const base = import.meta.env.BASE_URL.replace(/\/$/, '');
...
+      };
+    }
+
+    refresh(): void {
+      const input = this.dialog.querySelector<HTMLInputElement>('.pagefind-ui__search-input');
+      if (input && !this.wired) {
+        this.input = input;
+        input.setAttribute('role', 'combobox');
+        input.setAttribute('aria-autocomplete', 'list');
+        input.setAttribute('aria-controls', this.resultsContainerId);
```
JamesNK, line 122: **Nit:** Between the initial wiring here and the first Pagefind render (when `refresh()` sets `aria-expanded`), the input has `role="combobox"` but no `aria-expanded` attribute. Consider adding `input.setAttribute('aria-expanded', 'false');` in this block so the attribute is always present once the combobox role is set.

src/frontend/src/components/starlight/Search.astro
```diff
@@ -125,6 +467,138 @@ const base = import.meta.env.BASE_URL.replace(/\/$/, '');
...
+    display: flex;
+    flex-wrap: wrap;
+    align-items: center;
+    justify-content: center;
+    gap: 0.5rem 1.25rem;
+    margin-block-start: auto;
+    margin-inline: -1rem;
+    margin-block-end: -1rem;
+    padding: 0.6rem 1rem;
+    border-top: 1px solid var(--sl-color-gray-5);
```
JamesNK, line 565: **Nit:** These negative margins (`-1rem` / `-1.5rem`) and `bottom` values are coupled to Starlight's internal `.dialog-frame` padding. If Starlight ever changes that padding, this footer layout will break. Might be worth a comment noting the dependency so a future Starlight upgrade doesn't silently misalign the footer.

JamesNK (commented):
Nice work overall — clean separation of concerns, proper teardown, good a11y wiring, and solid e2e coverage.

The main issue is that **Home/End keys break text editing in the search input** since the capture-phase handler intercepts them regardless of focus target. The `aria-selected` on non-`option` elements is a smaller correctness fix. The rest are nits.

Architecture, RAF coalescing, `encodeURIComponent` for query params, `prefers-reduced-motion` support, and the `data-api-base-href` pattern are all well done.

src/frontend/src/components/starlight/Search.astro
```diff
@@ -17,52 +19,273 @@ const base = import.meta.env.BASE_URL.replace(/\/$/, '');
...
+    refresh(): void {
+      const input = this.dialog.querySelector<HTMLInputElement>('.pagefind-ui__search-input');
+      if (input && !this.wired) {
+        this.input = input;
+        input.setAttribute('role', 'combobox');
+        input.setAttribute('aria-autocomplete', 'list');
+        input.setAttribute('aria-controls', this.resultsContainerId);
+        input.setAttribute('aria-expanded', 'true');
+        input.addEventListener('keydown', this.keydownHandler);
+        this.wired = true;
```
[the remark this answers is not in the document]
IEvangelist: Resolved in commit dbd2575f (round 2): a `dialogStateHandler` flips `aria-expanded` to `false` on the dialog's `close` event, and `refresh()` re-reads `dialog.open` on every render. Also (in 0d81762d) the initial wiring now seeds `aria-expanded='false'` so the attribute is always present once `role='combobox'` is set, even before the first Pagefind render. The Escape-closes-dialog test asserts the post-close `false` value.

src/frontend/src/components/starlight/Search.astro
```diff
@@ -17,52 +19,273 @@ const base = import.meta.env.BASE_URL.replace(/\/$/, '');
...
+      }
+
+      const resultsContainer = this.dialog.querySelector<HTMLElement>('.pagefind-ui__results');
+      if (resultsContainer) {
+        if (!resultsContainer.id) resultsContainer.id = this.resultsContainerId;
+        resultsContainer.setAttribute('role', 'listbox');
+      }
```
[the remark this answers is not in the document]
IEvangelist: Resolved in commit dbd2575f (round 2): `refresh()` now force-sets `resultsContainer.id = this.resultsContainerId` on every render, so the `aria-controls` reference is never broken by an upstream id assignment.

src/frontend/src/components/starlight/Search.astro
IEvangelist: Fixed in 0d81762d. `Home` and `End` are now gated by `e.target !== this.input` -- they only drive listbox navigation once focus has moved out of the search field, so caret-to-start / caret-to-end work as expected while typing. Added an e2e test (`Home/End in the search input do not focus the API/load-more buttons`) to lock the behaviour in.

src/frontend/src/components/starlight/Search.astro
IEvangelist: Fixed in 0d81762d. `aria-selected` is now only written when `el.getAttribute('role') === 'option'` -- both in `setActive()` (active + cleared paths) and in the `refresh()` stale-cleanup loop. The load-more / API buttons no longer get a listbox-style selected state.

src/frontend/src/components/starlight/Search.astro
IEvangelist: Fixed in 0d81762d. The initial wiring block now sets `input.setAttribute('aria-expanded', 'false')` right after `role='combobox'` / `aria-controls`, so the attribute is always present from the moment combobox semantics are applied. The line below it continues to refresh the value to match `dialog.open` on every render.

src/frontend/src/components/starlight/Search.astro
IEvangelist: Done in 0d81762d -- added an explanatory comment block above `.search-keyboard-footer` noting that the negative `margin-inline` / `margin-block-end` and matching sticky `bottom` are bleeding the footer back through Starlight's `.dialog-frame` padding (`1rem` / `1.5rem` at >= 50rem). If those tokens change in a future Starlight upgrade, the values here need to track them.

JamesNK (approved):
I haven't tested. I trust you'll exerise edge case scenarios to find bugs.

PR #917, opened 2026-05-11
Update Aspire AppHost SDK and packages to 13.3.0

eerhardt:
Updates the `Aspire.Dev.AppHost` project from Aspire 13.2.4 to 13.3.0:

- `Aspire.AppHost.Sdk` 13.2.4 → 13.3.0
- `Aspire.Hosting.Azure.AppService` 13.2.4 → 13.3.0
- `Aspire.Hosting.JavaScript` 13.2.4 → 13.3.0

Build and restore verified locally.

PR #912, opened 2026-05-11
Fixed outdated 14 Japanese documentations

takashiuesaka:
fixed #911

PR #919, opened 2026-05-12
Strip twoslash annotations from llms*.txt outputs

IEvangelist:
## Summary

Aspire's docs site recently adopted [`expressive-code-twoslash`](https://www.npmjs.com/package/expressive-code-twoslash) to enrich on-site TS/C# code blocks with hover popovers, completion menus, and inline error boxes. The plugin injects extra HTML into each code block — e.g. a `<div class="twoslash-popup-container">…</div>` sibling of the original token — which is great for human readers but flows through `starlight-llms-txt`'s `rehype-remark` pipeline and ends up serialized into the code blocks shipped to LLM tooling via `llms.txt`, `llms-full.txt`, and `llms-small.txt`:

````text
const builder`const builder: HostBuilder` = await createBuilder();
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ popup contents leaked
````

That breaks the contract LLM consumers reasonably expect: code blocks in `llms*.txt` should contain only the source the author wrote.

## Why the existing options don't solve it

- **`minify.customSelectors`** is the right *mechanism* (it uses `hast-util-select`), but it's gated on `file.data.starlightLlmsTxt.minify === true`, i.e. only `llms-small.txt`. `llms.txt`, `llms-full.txt`, and any `customSets` still leak.
- **`rawContent: true`** bypasses MDX rendering altogether, which isn't acceptable for a site that uses Starlight components (`<Tabs>`, `<Aside>`, `<FileTree>`, …) — the raw JSX would leak into the Markdown instead.
- **A global Astro `rehypePlugins` entry** would strip the markup from the rendered site itself, defeating the on-site UX the plugin was added for.

## Change

This PR layers a new top-level `removeSelectors: string[]` option onto `starlight-llms-txt@0.8.1` via the existing `pnpm patch` infrastructure introduced in #899, then wires the twoslash annotation class names into `astro.config.mjs`.

`removeSelectors` is added as a new unified step in `htmlToMarkdownPipeline`, inserted right after `rehypeParse` and before `minifyLlmsTxt`. It runs unconditionally, so it applies to **every** generated bundle:

```js
const htmlToMarkdownPipeline = unified()
  .use(rehypeParse, { fragment: true })
  .use(function removeSelectorsLlmsTxt() {
    return (tree) => {
      if (removeSelectors.length === 0) return;
      remove(tree, (_node) => {
        const node = _node as RootContent;
        for (const selector of removeSelectors) {
          if (matches(selector, node)) return true;
        }
        return false;
      });
    };
  })
  .use(function minifyLlmsTxt() { /* unchanged */ })
  …;
```

aspire.dev's `astro.config.mjs` opts in with the 5 annotation classes twoslash injects:

```ts
starlightLlmsTxt({
  …,
  removeSelectors: [
    '.twoslash-popup-container',
    '.twoslash-static',
    '.twoslash-completion',
    '.twoslash-error-box',
    '.twoslash-custom-box',
  ],
  …
}),
```

### Why these selectors and not the wrappers

Twoslash builds its hover UI by injecting popup `<div>`s as **siblings** of the original token inside a wrapper:

```html
<span class="twoslash-hover">
  <div class="twoslash-popup-container not-content">…popup…</div>
  builder                                        ← the original token, sibling of the popup div
</span>
```

`unist-util-remove` removes only the matched node and its subtree, never its siblings. So removing `.twoslash-popup-container` strips the popup contents and leaves the token text in place. The wrappers `.twoslash`, `.twoslash-hover`, `.twoslash-error-underline`, and `.twoslash-noline` are **deliberately not** in the list — adding any of them would also nuke the underlying source. A dedicated wrapper-class guard test makes that constraint explicit.

## Upstream

The same `removeSelectors` change is up at [delucis/starlight-llms-txt#105](https://github.com/delucis/starlight-llms-txt/pull/105). Once a release containing it ships, the two-hunk `removeSelectors` portion of this patch can be dropped and `starlight-llms-txt` bumped (small follow-up PR). The patch header in `src/frontend/patches/starlight-llms-txt@0.8.1.patch` links to that PR so the dependency hand-off is discoverable.

## Patch shape

The patch file is the same combined patch that #899 introduced, now extended with three additional hunks (no other changes to the existing `collapseCodeBlocks` hunks):

- `types.ts` — adds `removeSelectors?: string[]` to `StarlightLllmsTextOptions` with JSDoc + `@example`, and a matching non-optional field on `ProjectContext`.
- `index.ts` — passes `opts.removeSelectors ?? []` through `projectContext`.
- `entryToSimpleMarkdown.ts` — adds the `removeSelectorsLlmsTxt` unified step described above. The plugin function name shows up in unified stack traces.

Regenerated with `pnpm patch` so it applies cleanly as a single diff.

## Tests

`src/frontend/tests/unit/llms-txt-twoslash.vitest.test.ts` (7 cases, wired into the `test:unit` chain as `test:unit:llms-txt`):

1. **hover popup** — asserts the popup type signature and JSDoc are gone, but the original token (`builder`) survives verbatim.
2. **error annotation box** — asserts the `(2769) — No overload matches…` annotation is gone, but the offending source (`addProject('api', '../Api/Api.csproj', 'http')`) survives.
3. **custom-tag callout** — asserts the callout body is gone, the wrapped expression survives.
4. **`^?` static popup** — asserts the static popup body is gone.
5. **`^|` completion popup** — asserts the completion menu is gone.
6. **sanity check** — same fixture without `removeSelectors` configured; asserts the popup text DOES leak, proving the regression test is wired up correctly.
7. **wrapper-class guard** — pins the constraint that `.twoslash`, `.twoslash-hover`, and `.twoslash-error-underline` are not in any production selector list. If a future contributor naïvely adds a wrapper class to `removeSelectors`, this fails immediately.

The existing `test:unit:llms-small-whitespace` sentinel from #899 still passes — the combined patch keeps every property it asserts on (regex literal, branch structure, etc.).

## Pin

Already pinned to exact `0.8.1` by #899; no change needed here. Adding to the patch keeps that pin honest — a future range-bump would silently drop both patches and re-introduce both bugs.

## Verified

- `pnpm lint` — clean.
- `pnpm test:unit` — full chain green, including:
  - `test:unit:llms-small-whitespace` — 14/14 (#899's existing coverage).
  - `test:unit:llms-txt` — 7/7 (new).
- Local build behavior previously verified end-to-end: the popup type signatures / error annotations / completion menus disappear from `dist/llms.txt`, `dist/llms-full.txt`, and `dist/llms-small.txt`, while the original code (e.g. `'http'`, `'https'`, `builder`, `createBuilder()`) is preserved verbatim. On-site Twoslash UX is unaffected.

## Out of scope

- Removing the `removeSelectors` hunk after the upstream release ships (small follow-up PR).
- Renaming the option if the upstream maintainer prefers a different name during review of delucis/starlight-llms-txt#105 — the local call site is one place to fix up.


IEvangelist:
Thanks for catching this @copilot-pull-request-reviewer — really good point. The inlined pipeline was effectively only testing my reimplementation of the algorithm, not the patch or the call site.

Fixed in bfc0d400b79c32b979ab3455421e86482f2aebaf by adding two sentinel `describe` blocks, modelled on the `llms-small-whitespace.vitest.test.ts` pattern:

1. **Patch sentinel** — reads the actual patched files in `node_modules/starlight-llms-txt/`:
   - `entryToSimpleMarkdown.ts` must contain the new `.use(function removeSelectorsLlmsTxt(...))` step, read `starlightLllmsTxtContext.removeSelectors`, call `matches(selector, node)`, and short-circuit on `removeSelectors.length === 0`.
   - `types.ts` must declare `removeSelectors?: string[]` on the public options and `removeSelectors: NonNullable<StarlightLllmsTextOptions[''removeSelectors'']>` on `ProjectContext`.
   - `index.ts` must forward `opts.removeSelectors ?? []`.
2. **Wiring sentinel** — reads `astro.config.mjs`, extracts the `removeSelectors: [...]` array passed to `starlightLlmsTxt({...})`, and asserts the parsed selector list equals (as a set) the `REMOVE_SELECTORS` constant under test. So if the call site is removed, or if someone adds/removes a selector in either place without updating the other, this fires.

Verified locally:

- All 11 tests pass when the patch and call site are intact.
- Negative test: renaming `function removeSelectorsLlmsTxt(` to `function someOtherName(` in the patched `node_modules/starlight-llms-txt/entryToSimpleMarkdown.ts` correctly fails the patch sentinel with:
  ```
  AssertionError: patched plugin should register a removeSelectorsLlmsTxt unified step:
    expected '...' to match /\.use\(function removeSelectorsLlmsTxt\(/
  ```

Going to stop short of "invoke the patched function directly" since `entryToSimpleMarkdown.ts` pulls in `experimental_AstroContainer`, `astro:content`, `virtual:starlight-llms-txt/context`, etc., which would require booting the Astro/Starlight runtime in a unit test. The sentinel-style coverage is consistent with how PR #899 handles the same kind of patch.

mitchdenny (approved):
The approach is well-considered — the patch is scoped tightly, the version is pinned exactly, tests include sentinels that will catch patch drift, and the same change is already up for review upstream at delucis/starlight-llms-txt#105 with a clear follow-up plan to drop the patch hunk once it ships. LGTM.

src/frontend/patches/starlight-llms-txt@0.8.1.patch
```diff
@@ -10,7 +10,40 @@ index 5fdc6c54ae6edc5874db921dde4b89ed9e15dc3c..c9561aadb8ec4be885a828b7a5ebc2da
  	customSelectors: [],
  };
  /** Resolved minification options */
-@@ -192,7 +193,31 @@ export async function entryToSimpleMarkdown(
+@@ -28,12 +29,32 @@ const minify = { ...minifyDefaults, ...starlightLllmsTxtContext.minify };
+ const selectors = [...minify.customSelectors];
```
JamesNK: Are you going to upstream this feature to the library's repo?

IEvangelist:
Followed @delucis''s feedback on the upstream PR ([delucis/starlight-llms-txt#105](https://github.com/delucis/starlight-llms-txt/pull/105)) and reworked the API there. To keep the two PRs in sync, pushed 2c9b7ccd here:

- The upstream PR no longer introduces a `removeSelectors` option. Instead it extends the existing `customSelectors` option to accept a per-output object shape (`{ small?: string[]; full?: string[]; all?: string[] }`), while still accepting the legacy `string[]` form (which continues to apply to `llms-small.txt` only). The deprecated `minify.customSelectors` option keeps working and is merged additively into the `small` bucket.

- `patches/starlight-llms-txt@0.8.1.patch` was regenerated via `pnpm patch` against the new upstream diff. The combined patch still contains @LuceCarter''s `collapseCodeBlocks` work from #899 — only my hunks changed.

- `astro.config.mjs` switched from `removeSelectors: [...]` to:

  ```ts
  customSelectors: {
    all: [
      ''.twoslash-popup-container'',
      ''.twoslash-static'',
      ''.twoslash-completion'',
      ''.twoslash-error-box'',
      ''.twoslash-custom-box'',
    ],
  },
  ```

  `all` merges into both the small and full buckets, so the twoslash annotations are still stripped from every output (`llms.txt`, `llms-full.txt`, `llms-small.txt`, and any `customSets`) — same end behavior as before.

- The sentinel tests were updated to pin the new symbols (`customSelectorsLlmsTxt`, `smallSelectors`, `fullSelectors`, the `file.data...minify` ternary) and the new `customSelectors: { all: [...] }` wiring shape. Verified with positive runs and two negative tests (renaming the patched function / renaming the config key) — each correctly fails the matching sentinel.

Local checks all green:

| Check | Result |
|---|---|
| `pnpm test:unit:llms-txt` | 11/11 pass |
| `pnpm test:unit:llms-small-whitespace` (from #899) | 14/14 pass |
| `pnpm lint` | clean |
| Negative test (patch sentinel) | correctly fails when patched function is renamed |
| Negative test (wiring sentinel) | correctly fails when `customSelectors` key is renamed in `astro.config.mjs` |

Title still reads "removeSelectors" — happy to also rename the PR title here once upstream lands. For now I left it so the cross-PR discussion stays threaded.

IEvangelist:
Rebased onto latest `upstream/main` (tip `428f9f98`) to resolve the conflicts with #884, #899, #910, #912, #916, #917, #922, #924.

**Resolved conflicts**
- `src/frontend/package.json` — kept the dependabot bumps from #924 (`@playwright/test`, `@astrojs/starlight`, `astro`, `typescript-eslint`, `vitest`); preserved this PR's new devDependencies (`@types/hast`, `hast-util-select`, `rehype-parse`, `rehype-remark`, `remark-gfm`, `remark-stringify`, `unified`, `unist-util-remove`) and the `test:unit:llms-txt` script.
- `src/frontend/pnpm-lock.yaml` — took `upstream/main` and regenerated via `pnpm install` so the lockfile reflects both the upstream bumps and this PR's new deps. Patch hash for `starlight-llms-txt@0.8.1` re-fingerprinted automatically.

**Local verification (post-rebase)**
- `pnpm install` — clean (no new peer warnings beyond what `main` already emits).
- `pnpm test:unit:llms-txt` — 11/11 ✓
- `pnpm test:unit:llms-small-whitespace` — 14/14 ✓ (PR #899's tests still green)
- `pnpm lint` — clean.

No conflicts with the Japanese `aspire-13-2.mdx` translation (#922) or the inline-code CSS work (#884) since those touch unrelated files.

New tip: `39777acf`.

src/frontend/patches/starlight-llms-txt@0.8.1.patch
IEvangelist: Yes, https://github.com/delucis/starlight-llms-txt/pull/105

PR #941, opened 2026-05-13
ci: Phase 1 low-risk CI wall-time optimizations

IEvangelist:
## Summary

Phase 1 of CI wall-time optimizations — all changes are low-risk with no behavioural trade-offs. Each item is independently revertable. Background research and tradeoffs are documented in the commit message.

### Frontend / Vitest
- **Consolidate `test:unit`** — was 10 sequential `vitest run` calls, each paying full Vite + TS-transform cold-start cost. Now a single invocation that picks up the same files via `vitest.config.ts`'s `include` pattern. Individual `test:unit:*` scripts kept for local granular runs.
- **Vitest `pool: 'threads'`** — `worker_threads` has lower spawn overhead than the default `forks` pool on Linux. Tests are `environment: 'node'` with no native addons or thread-unsafe APIs.
- **Cache `node_modules/.astro`** keyed on content + `astro.config.mjs` so Astro's content layer doesn't re-hydrate from scratch when MDX hasn't changed.
- **`pnpm install --prefer-offline`** — skips registry HEAD probes when the setup-node pnpm-store cache hits. Falls back to network on misses.
- **`compression-level: 0` on Playwright artifact uploads** — reports contain PNGs, webm videos, and traces that are already compressed.

### .NET
- **`DOTNET_NOLOGO=1` and `DOTNET_CLI_TELEMETRY_OPTOUT=1`** on both `apphost-build.yml` and `tools-tests.yml`.
- **`tools-tests.yml` explicit build + `--no-build`** — `dotnet test` was implicitly rebuilding each project. Now we restore once, build once, and run `dotnet test --no-build --no-restore`.

### GitHub Actions infra
- **`concurrency` group on `ci.yml`** with `cancel-in-progress: ${{ github.event_name == 'pull_request' }}`. Stacked PR runs cancel after rebases; pushes to `main`/`release/*` are never cancelled.
- **`changes` job uses `fetch-depth: 1`** plus a targeted `git fetch --depth=1` of the PR base and head SHAs. `frontend-build.yml` keeps `fetch-depth: 0` because Lunaria still requires full history.

## Verified locally

- `pnpm test:unit` runs all **159 unit tests across 17 files in 5.5 s** with the new consolidated invocation + threads pool.
- YAML for all four workflows parses cleanly.

## Explicitly NOT recommended (researched and verified)

Listed here so reviewers don't ask:
- `NODE_OPTIONS=--max-old-space-size=*` — Node 22+ auto-sizes heap to ~50% of RAM; manual flags are cargo cult on 16 GB runners.
- `pnpm --reporter=silent` — already silent in non-TTY CI.
- `DOTNET_SKIP_FIRST_TIME_EXPERIENCE` — deprecated since .NET Core 3.0.
- `build:skip-search` — Pagefind IS exercised by `site-search.spec.ts` and `ts-api-search.spec.ts`.
- Vitest `isolate: false` — `api-markdown.vitest.test.ts` uses `vi.mock`; state-bleed risk isn't worth the marginal win.
- Caching `~/.cache/ms-playwright` — Playwright's own docs advise against it.

## Follow-ups (not in this PR)

Two higher-reward changes that trade off discipline or runner minutes; I'll open separate PRs once we agree:

- **NuGet caching with `packages.lock.json`** — adds dependency-bump discipline.
- **Shard Playwright across viewport projects** (`desktop-chromium`, `tablet-chromium`, `mobile-chromium`) — 3× runner minutes for the E2E phase in exchange for ~60% wall-time reduction.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,11 +6,17 @@ on:
   pull_request:
     branches: [ main, release/* ]
 
-# Cancel superseded runs on the same PR branch. Pushes to main/release/* are
-# never cancelled — branch protection still needs the gate to pass on the
-# latest commit.
+# Concurrency strategy:
+# - On pull_request: group by PR number. New commits to the PR cancel the
+#   in-flight run for the same PR.
+# - On push (main/release/*): include `github.run_id` so every push gets a
+#   unique group. This guarantees pushes never cancel each other, even if
+#   several commits land while an earlier run is still in progress.
+#   (With a shared group, GHA cancels older *pending* runs even when
+#   `cancel-in-progress: false`, which would silently drop required CI on
+#   `main`.)
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 permissions:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,13 @@ on:
   pull_request:
     branches: [ main, release/* ]
 
+# Cancel superseded runs on the same PR branch. Pushes to main/release/* are
+# never cancelled — branch protection still needs the gate to pass on the
+# latest commit.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 permissions:
   contents: read
   security-events: write
@@ -19,7 +26,18 @@ jobs:
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
-          fetch-depth: 0
+          # Shallow clone; the two PR SHAs we need to diff against are
+          # fetched explicitly below. This is much cheaper than `fetch-depth: 0`.
+          fetch-depth: 1
+
+      - name: Fetch PR base and head for diff
+        if: github.event_name == 'pull_request'
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+        run: |
+          git fetch --depth=1 origin "$BASE_SHA"
+          git fetch --depth=1 origin "$HEAD_SHA"
 
       - id: filter
         name: Detect changed areas
--- a/.github/workflows/frontend-build.yml
+++ b/.github/workflows/frontend-build.yml
@@ -63,7 +63,18 @@ jobs:
           cache-dependency-path: src/frontend/pnpm-lock.yaml
 
       - name: Install deps
-        run: pnpm install --frozen-lockfile
+        # `--prefer-offline` skips the registry staleness HEAD-request when
+        # the setup-node pnpm-store cache hits. pnpm falls back to network
+        # for anything missing, so cache misses still work.
+        run: pnpm install --frozen-lockfile --prefer-offline
+
+      - name: Cache Astro content layer
+        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
+        with:
+          path: src/frontend/node_modules/.astro
+          key: astro-content-${{ runner.os }}-${{ hashFiles('src/frontend/src/content/**', 'src/frontend/src/content.config.*', 'src/frontend/astro.config.mjs') }}
+          restore-keys: |
+            astro-content-${{ runner.os }}-
 
       # Build BEFORE tests so that Lunaria (i18n status) runs against a
       # pristine working tree.  The lint step inside test:all calls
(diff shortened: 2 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -6,6 +6,13 @@ on:
   pull_request:
     branches: [ main, release/* ]
 
+# Cancel superseded runs on the same PR branch. Pushes to main/release/* are
+# never cancelled — branch protection still needs the gate to pass on the
+# latest commit.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
```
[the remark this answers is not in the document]
IEvangelist: Good catch — you're right that GHA cancels older *pending* runs in the same group even when `cancel-in-progress: false`, which could drop a required `main` run if commits land in quick succession.

Fixed in 43b4669c by switching the group key to include `github.run_id` for non-PR events:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

- `pull_request`: group keyed on PR number — new commits cancel in-flight runs for the same PR (intended).
- `push`: `run_id` makes every push its own group, so pushes never cancel/queue against each other.

.github/workflows/ci.yml
```diff
@@ -6,6 +6,19 @@ on:
...
+# - On pull_request: group by PR number. New commits to the PR cancel the
+#   in-flight run for the same PR.
+# - On push (main/release/*): include `github.run_id` so every push gets a
+#   unique group. This guarantees pushes never cancel each other, even if
+#   several commits land while an earlier run is still in progress.
+#   (With a shared group, GHA cancels older *pending* runs even when
+#   `cancel-in-progress: false`, which would silently drop required CI on
+#   `main`.)
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
```
eerhardt, line 19: Hmm, why isn't this the default behavior? I don't understand this change.

.github/workflows/frontend-build.yml
```diff
@@ -63,7 +63,18 @@ jobs:
           cache-dependency-path: src/frontend/pnpm-lock.yaml
 
       - name: Install deps
-        run: pnpm install --frozen-lockfile
+        # `--prefer-offline` skips the registry staleness HEAD-request when
+        # the setup-node pnpm-store cache hits. pnpm falls back to network
+        # for anything missing, so cache misses still work.
+        run: pnpm install --frozen-lockfile --prefer-offline
+
+      - name: Cache Astro content layer
```
eerhardt, line 71: Is this safe to do? What are the risks?

.github/workflows/ci.yml
IEvangelist: Good question. Two things going on:

**Why GHA doesn't do this by default:** concurrency control isn't universally safe — release pipelines, audit jobs, and workflows with external side-effects need every commit to run to completion. So GHA's default is "run everything in parallel" and `concurrency:` is the opt-in switch.

**What this specific block does:**

- **PRs** — key includes `github.event.pull_request.number` + `cancel-in-progress: true`. New commits to a PR cancel the older in-flight run for that same PR. Saves runner-minutes on rapid fixup / rebase pushes.
- **Pushes (`main` / `release/*`)** — key falls through to `github.run_id`, which is unique per run, so each push effectively gets its own group. Combined with `cancel-in-progress: false`, push runs neither cancel each other nor queue against each other. Every commit to `main` always gets its own CI run.

The reason this is split rather than just using a single shared group with `cancel-in-progress: false` is the bug the bot caught above: when runs share a group, GHA cancels older *pending* runs regardless of `cancel-in-progress`. Giving push events a unique group sidesteps that entirely.

.github/workflows/frontend-build.yml
IEvangelist: Safe — worst case is a cold rebuild.

**What this is:** `node_modules/.astro` is Astro's content-layer cache — pre-validated MDX entries, generated content-collection types, and incremental state created during `astro sync` / `astro build`. Caching it across runs lets Astro skip re-parsing and re-validating MDX that hasn't changed.

**Risks:**

1. **Stale cache** — addressed by the key: `hashFiles('src/content/**', 'src/content.config.*', 'astro.config.mjs')`. Any change to a content file, the content config, or the Astro config invalidates it. Astro also re-validates entries on load, so even if a stale match restored, it would self-correct.
2. **Astro version bumps** — deliberately not in the key. If Astro changes its cache format the worst case is one cold build while Astro regenerates incompatible entries. Adding `pnpm-lock.yaml` to the key would invalidate on every dep bump, which is way more frequent than format changes.
3. **Cross-branch pollution** — GHA caches are scoped per-branch with a fallback to the default branch. A PR can only restore caches from its own branch or `main`. No cross-PR contamination.
4. **Cache restore failure** — falls through to a cold build. No path where this produces an incorrect build.

If you'd prefer the belt-and-suspenders version (include `pnpm-lock.yaml` in the key so Astro upgrades are always cold), happy to add it — accepts more cache misses in exchange for a stricter invalidation guarantee.

.github/workflows/ci.yml
eerhardt: > New commits to a PR cancel the older in-flight run for that same PR. Saves runner-minutes on rapid fixup / rebase pushes.

I thought that is the normal "default" behavior...

.github/workflows/frontend-build.yml
eerhardt: sounds good.

eerhardt (approved):

PR #950, opened 2026-05-14
Add first-class Go hosting docs

davidfowl:
## Summary
- Replace the deprecated CommunityToolkit Go hosting page with first-class `Aspire.Hosting.Go` guidance.
- Cover `AddGoApp` / `addGoApp`, app directories and package paths, app arguments, module helpers, Delve debugging, publish behavior, private modules, and experimental Go templates.
- Keep the PR scoped to `src/frontend/src/content/docs/integrations/frameworks/go-apps.mdx` on `release/13.4`.

## Validation
- `pnpm --dir .\src\frontend exec prettier --check src/content/docs/integrations/frameworks/go-apps.mdx`
- `git diff --check`
- `pnpm --dir .\src\frontend run check-data`
- `pnpm --dir .\src\frontend run test:unit:docs`

Note: `pnpm --dir .\src\frontend run lint` currently fails on `release/13.4` in `tests/unit/llms-txt-twoslash.vitest.test.ts` with existing `@typescript-eslint/no-unsafe-*` errors unrelated to this docs page.


davidfowl:
cc @edmondshtogu

IEvangelist (commented):
While I understand that this is a replacement, we do need to restructure this so that it follows the new pattern we recently implemented. Ideally, this will have a sidebar like:

```
Go
- Get started
- Set up Go in the AppHost
```

The get started should answer, "what (intro p)", "why (from broad to aspire specficially)", "how (the pieces fit together)", and then step links out to Modeling Go in the AppHost. And it might be a good idea to even show Connect to Go from a consuming resource, which would explain injected config from env var and how to use them - connection properties, etc.

davidfowl:
We need to merge main into release/13.4 first.

IEvangelist:
> We need to merge main into release/13.4 first.

It's still a draft PR, do you want it merged now?

src/frontend/src/data/integration-docs.json
```diff
@@ -409,7 +413,7 @@
   },
   {
     "match": "CommunityToolkit.Aspire.Hosting.Golang",
-    "href": "/integrations/frameworks/go-apps/"
+    "href": "/integrations/frameworks/go/go-get-started/"
   },
```
IEvangelist: This should now be removed. And we need to add it to the exclusions for packages that are automatically pulled, add an exclude entry here:

https://github.com/microsoft/aspire.dev/blob/main/src/frontend/scripts/update-integrations.ts#L15-L25

For `CommunityToolkit.Aspire.Hosting.Golang`.

IEvangelist (approved):

src/frontend/src/data/integration-docs.json
davidfowl: Accepted. Removed the deprecated `CommunityToolkit.Aspire.Hosting.Golang` doc mapping and added it to `EXCLUDED_PACKAGES` so `update-integrations` will not re-add it.

PR #958, opened 2026-05-14
ci: Phase 2A — NuGet caching with packages.lock.json

IEvangelist:
# Phase 2A — NuGet caching with `packages.lock.json`

Follow-up to #941 (Phase 1, which delivered a ~9% wall-time reduction on the
critical `Frontend Build` job — see `dotnet test` measurements at the end).
This PR is Phase 2 item A from the original optimization plan.

## What this does

Enables `actions/setup-dotnet` package caching for the two .NET workflows
(`apphost-build.yml` and `tools-tests.yml`). The action caches the global
`~/.nuget/packages` folder, keyed on a hash of each project's
`packages.lock.json`. On a cache hit, `dotnet restore --locked-mode` becomes
a manifest verification only — no network calls to nuget.org.

## Changes

- Generated `packages.lock.json` for the six .NET projects that participate
  in CI restore (`Aspire.Dev.AppHost`, `StaticHost`, `PackageJsonGenerator`,
  `AtsJsonGenerator`, and the two test projects).
- Added `<RestorePackagesWithLockFile>true</RestorePackagesWithLockFile>`
  to each csproj so the lock files are honored by default. `--locked-mode`
  then makes CI fail loudly if anything ever drifts.
- `apphost-build.yml` and `tools-tests.yml`:
  - `cache: true` + `cache-dependency-path` on `actions/setup-dotnet`.
  - Switched `dotnet restore` to `--locked-mode`.

## Expected impact

`actions/setup-dotnet` caching typically shaves 15–60 s off each .NET job
on a cache hit. The current `AppHost Build` job is small (~35 s end-to-end,
~8 s in the restore step), so the absolute win is modest. The bigger value
is **predictable restore time and an integrity guarantee** — `--locked-mode`
means a transitive dependency can't silently change between builds.

## Contributor impact

When bumping any `PackageReference`, regenerate the lock file:

```bash
dotnet restore <project>.csproj --use-lock-file --force-evaluate
```

Commit the updated `packages.lock.json` alongside the version change. If
you forget, CI fails with a clear "Lock file is out of sync" error.

## Verification

Verified locally that `dotnet restore --locked-mode` succeeds for all three
CI entry points (`AppHost`, `PackageJsonGenerator.Tests`,
`AtsJsonGenerator.Tests`), and that `dotnet build --no-restore` succeeds
for each. No package-feed changes (still only `nuget.org`).

### DIFF
--- a/.github/workflows/apphost-build.yml
+++ b/.github/workflows/apphost-build.yml
@@ -19,9 +19,13 @@ jobs:
     - uses: actions/setup-dotnet@c2fa09f4bde5ebb9d1777cf28262a3eb3db3ced7 # v5.2.0
       with:
         global-json-file: global.json
+        cache: true
+        cache-dependency-path: |
+          src/apphost/Aspire.Dev.AppHost/packages.lock.json
+          src/statichost/StaticHost/packages.lock.json
 
     - name: Restore
-      run: cd src/apphost/Aspire.Dev.AppHost && dotnet restore
+      run: cd src/apphost/Aspire.Dev.AppHost && dotnet restore --locked-mode
 
     - name: Build
       run: cd src/apphost/Aspire.Dev.AppHost && dotnet build --no-restore --configuration Release
--- a/.github/workflows/tools-tests.yml
+++ b/.github/workflows/tools-tests.yml
@@ -33,11 +33,17 @@ jobs:
       - uses: actions/setup-dotnet@c2fa09f4bde5ebb9d1777cf28262a3eb3db3ced7 # v5.2.0
         with:
           global-json-file: global.json
+          cache: true
+          cache-dependency-path: |
+            tests/PackageJsonGenerator.Tests/packages.lock.json
+            tests/AtsJsonGenerator.Tests/packages.lock.json
+            src/tools/PackageJsonGenerator/packages.lock.json
+            src/tools/AtsJsonGenerator/packages.lock.json
 
       - name: Restore tests
         run: |
-          dotnet restore tests/PackageJsonGenerator.Tests/PackageJsonGenerator.Tests.csproj
-          dotnet restore tests/AtsJsonGenerator.Tests/AtsJsonGenerator.Tests.csproj
+          dotnet restore tests/PackageJsonGenerator.Tests/PackageJsonGenerator.Tests.csproj --locked-mode
+          dotnet restore tests/AtsJsonGenerator.Tests/AtsJsonGenerator.Tests.csproj --locked-mode
 
       - name: Build tests
         run: |
--- a/src/apphost/Aspire.Dev.AppHost/Aspire.Dev.AppHost.csproj
+++ b/src/apphost/Aspire.Dev.AppHost/Aspire.Dev.AppHost.csproj
@@ -6,6 +6,7 @@
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
     <UserSecretsId>c2fee4fa-ede6-4b31-88bf-7f029684ef8d</UserSecretsId>
+    <RestorePackagesWithLockFile>true</RestorePackagesWithLockFile>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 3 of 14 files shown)


IEvangelist:
## Wall-time measurement — converting to draft

After running CI twice on this branch (once cold, once with a primed cache), the
NuGet caching turns out to be roughly a wash on this repo. The .NET projects
here are small enough that the cache restore / cache save overhead cancels out
the savings from skipping the package downloads.

### `apphost-build` (AppHost Build)

| Run                       | Job total | `setup-dotnet` | `Restore` | `Post setup-dotnet` |
|---------------------------|----------:|---------------:|----------:|--------------------:|
| `main` baseline (no cache, no lock file) — median ≈ | **32 s** | 7 s | 8 s | 1 s |
| This PR, run 1 (cache miss + save) | **59 s** | 8 s | 14 s | 10 s |
| This PR, run 2 (cache hit) | **33 s** | 14 s | 2 s | 0 s |

The cache hit (run 2) is essentially identical to the no-cache baseline. The
Restore step drops from 8 s → 2 s, but `setup-dotnet` itself takes 7 s longer
to download and rehydrate the cache, so they cancel.

### Tools Tests

| Run                       | Job total | `setup-dotnet` | `Restore` |
|---------------------------|----------:|---------------:|----------:|
| `main` baseline (no cache, no lock file) — median ≈ | **34 s** | 7 s | 8 s |
| This PR, run 1 (cache miss + save) | **57 s** | 9 s | 9 s |
| This PR, run 2 (cache hit) | **44 s** | 4 s | 5 s |

Cache hit is ~10 s slower than baseline. The Restore step drops by 3 s but
that doesn't recover the ~8 s of extra checkout / build variance.

### Why the original estimate was wrong

The Phase 2 plan estimated "~15–60 s saved per .NET job". That number is
realistic for medium-sized .NET solutions where `dotnet restore` over the wire
takes 30–60 s. The AppHost + tools projects here resolve in ~8 s without a
cache — the speed-of-light floor is too close to the existing baseline for
caching to be a net win.

### Trade-offs

- ❌ **Wall time:** zero or slightly negative on this repo.
- ➖ **Maintenance:** every dependency bump now requires
  `dotnet restore --use-lock-file --force-evaluate` **on Linux** (Docker is
  fine). On Windows the hashes drift (NU1403); we hit this during this PR
  and had to regenerate two lock files inside a Linux container.
- ✅ **Supply-chain integrity:** lock files do give us reproducible restores
  and an audit trail for transitive package versions. That's a real benefit,
  just not the one this PR was scoped for.

### Recommendation

Converting this PR to **draft**. If we later want lock files for supply-chain
reasons, this work can be picked back up. If we just want the CI-speed angle,
this is not the lever to pull on this repo — the .NET jobs are already at
their fast-path floor.

The companion PR (https://github.com/microsoft/aspire.dev/pull/959, Playwright
sharding) is the high-signal Phase 2 item — see findings there.

adamint (approved):

IEvangelist:
Following up on the pushback that Phase 2A might not be a real perf win — I pulled the post-merge numbers and the data agrees. Opened a revert in #964 with the full comparison.

TL;DR from the comparison:

| job                   | n (pre) | n (post) | pre median | post median | net |
| --------------------- | ------- | -------- | ---------- | ----------- | --- |
| `apphost-build`       | 4       | 5        | 34 s       | 33 s        | -1 s (-2.9 %, within noise) |
| `Tools Tests`         | 5       | 1        | 35 s       | 44 s        | +9 s (cache miss/populate) |

`dotnet restore` on these projects already runs in ~8 s without a cache; the `setup-dotnet` cache-restore step itself adds 7–8 s, so the savings cancel. Variance widened too (best 24 s, worst 53 s post-merge).

Combined with the lock-file maintenance friction (Linux-only regen to avoid NU1403, which we hit on this PR itself), there's no perf justification to keep this. Phase 2B (#959) is the Phase 2 win that's actually worth keeping.

The supply-chain angle for `packages.lock.json` remains — if we want it for reproducible restores / transitive-dep audit trail, happy to revisit as a focused change with that as the explicit goal.

Issue #913, opened 2026-05-11
Translation Skill for Documentation Localization


takashiuesaka:
I have manually translated several documents into Japanese so far, but it has been taking too much time. So, I created a translation skill and tried using it to retranslate Japanese documents that had become outdated, and I was able to get very good results. The skill includes the knowledge and experience I accumulated while translating documents manually.

I designed the skill so that it could be used with other languages as much as possible, although I have not tested it with languages other than Japanese.

Using this skill, it should be possible to continuously generate newly translated documents with GitHub Actions. I would like to provide this skill so that others can make use of it.

IEvangelist:
Is this skill generic, or specific to Japanese?

takashiuesaka:
I am creating the skill to be generic, but I can only test Japanese translation because I only understand Japanese and English.

takashiuesaka:
I am still working on the skill and checking the quality of the results. I realized that language-specific instructions are needed, so I created another markdown file for Japanese translation and referenced it from SKILL.md. So far, it has been working well.

IEvangelist:
Yeah, I was going to recommend references for each supported language. Looking forward to it. Thank you!

IEvangelist:
### Triage → @IEvangelist

This proposes a *new skill* (translation), not a doc fix. Decisions you'll need to make before Copilot can build anything:
- Does this skill live in `microsoft/aspire-skills` or in a separate i18n-focused repo?
- What's the ownership model for translated output (per-language reviewers, ja-JP only, etc.)?
- Relationship with the existing translation pipeline implied by the populated `ja/`, `de/`, `es/`, `fr/`, `hi/`, `id/`, `it/`, `ko/`, `pt-br/`, `ru/`, `tr/`, `uk/`, `zh-cn/`, `da/` doc trees.

Reach out to @takashiuesaka — they have a working draft of the skill.

PR #978, opened 2026-05-15
Translate documents into Japanese using "Skill"

### LINKED ISSUE #976: Translate documents into Japanese using "Skill"
As a result of creating and checking the quality of the "translator skill", many documents have been translated.
Of course, the quality is good.

takashiuesaka:
fixed #976 
related #913


takashiuesaka:
Hi @IEvangelist, could you please help me?

I got the following error during the test phase in GitHub Actions:

> Error: Found 1 unexpected diagnostic(s) in 1 block(s):
>
> ✖ ja/app-host/container-registry.mdx:341:33 (block #7) ts(2345) — Argument of type 'AzureContainerRegistryResource' is not assignable to parameter of type 'IResource'.

I checked the code block that caused the error in both the original file (/app-host/container-registry.mdx) and the translated file (/ja/app-host/container-registry.mdx), but they appear to be identical.

Do you have any idea what might be causing this error?

takashiuesaka:
OK, I found the reason why the error occurs. This issue is caused by twoslash. PR #940 fixed this issue, but the fix only applies to English.

I confirmed that the same error occurs in my local environment by running the pnpm test:unit:twoslash-blocks command.

IEvangelist (approved):

PR #1006, opened 2026-05-19
feat(gallery): hide client integrations, add "aspire add" CTA and language buttons

IEvangelist:
## Summary

Reworks the integrations gallery (`/integrations/gallery/`) along three axes:

1. **Hide client integrations** from the rendered grid behind a new `PUBLIC_SHOW_CLIENT_INTEGRATIONS` feature flag (defaults off), and remove the hosting/client toggle group entirely. The underlying `aspire-integrations.json` / `integration-docs.json` data files, generated API docs, and content articles continue to be generated and updated — only the gallery surface hides them.
2. **Add an `aspire add {shortName}` snippet** to every card (using the existing `generateFriendlyName` algorithm in `src/utils/helpers.ts`, which mirrors the Aspire CLI's `AddCommand`), with an inline copy-to-clipboard button.
3. **Add two full-width touching C# / TypeScript icon buttons** under each snippet linking to the integration's hosting docs page with a new global `?aspire-lang=csharp|typescript` query parameter. A `PivotSelector` whose `key === "aspire-lang"` now also bridges its selection into `localStorage['starlight-synced-tabs__aspire-lang']`, so any prior pivot choice defaults `<Tabs syncKey="aspire-lang">` on destination pages.

## What changed

- `src/components/Integrations.astro` — Reads `PUBLIC_SHOW_CLIENT_INTEGRATIONS`, filters client-tagged entries when off, and gates the hosting/client toggle group + its event listeners + URL-param writes.
- `src/components/IntegrationCard.astro` — Adds the `aspire add` snippet, copy button (delegated `click` handler via a hoisted Astro `<script>`), and the C# / TS button row. Drops the redundant "Docs" footer link in favour of the language buttons. Maps `*-get-started/` → `*-host/` for the button hrefs (that's where the `aspire-lang` `<Tabs>` actually live).
- `src/components/PivotSelector.astro` — Inside `apply(id)`, when `qsKey === 'aspire-lang'`, mirrors the selection into `starlight-synced-tabs__aspire-lang` using the same `{ csharp: 'C#', typescript: 'TypeScript' }` label map already used by `starlight/Head.astro`.
- `src/content/i18n/en.json` — New `integrations.addCommand`, `copyCommand`, `copyCommandCopied`, `viewLanguageDocs`, `viewCsharpDocs`, `viewTypescriptDocs` strings. Other locales fall back to English to keep translation churn low.
- `tests/e2e/integrations-gallery.spec.ts` (new) — Covers: client cards hidden + hosting/client toggle group absent (skipped when the flag is on), `aspire add postgresql` snippet present, copy button writes to clipboard and shows the copied state, C#/TS hrefs share the same base + carry the right `aspire-lang`, hrefs target the `*-host/` page (not `*-get-started/`), clicking C# pre-selects the C# tab + populates both storage keys, and the PivotSelector → synced-tabs bridge.

## Notes / decisions

- **Why `*-host/` instead of `*-get-started/`?** `integration-docs.json` maps titles to `*-get-started/` landing pages, but `<Tabs syncKey="aspire-lang">` only exists on the `*-host/` sibling. Without the swap, the language buttons would land on a page with no synced tabs — making the whole CTA cosmetic.
- **External docs (e.g. AWS)** are skipped for language buttons via a `supportsLangTabs = pkg.docs?.startsWith('/')` guard; the snippet still renders.
- **The hosting/client toggle group is removed when the flag is off** rather than just hiding the client cards but keeping the toggle — per the ask, the toggle UI goes away too.
- **`integrations.visitDocs` and `integrations.hosting* / client*` translation keys are intentionally left in place** to avoid translation churn. They're unused by the new card layout but will gracefully re-enter use if the flag is flipped back on.
- **Copy button uses event delegation on `document`** with a single hoisted `<script>` so we don't ship per-card listeners for 100+ cards. Includes a `document.execCommand('copy')` fallback for non-clipboard contexts.

## Validation

- `pnpm lint` ✅
- `pnpm test:unit` ✅ (220/220)
- `pnpm check-data` ✅
- `pnpm build:skip-search` ✅ (12,226 pages built)
- Branch rebased on top of latest `upstream/main` (`1cb508fd`) to avoid merge conflicts.

### DIFF
--- a/src/frontend/src/components/Integrations.astro
+++ b/src/frontend/src/components/Integrations.astro
@@ -199,19 +199,26 @@ const sortedIntegrations = visibleIntegrations.sort((a, b) => {
     display: none; /* Hidden by default on mobile */
   }
 
+  /* The card grid's column count is driven by a container query against the
+     `.equal-columns` wrapper below — viewport media queries can't see the
+     Starlight sidebar carving ~17rem out of the available width, so they
+     flipped the grid to 2 columns when the actual gallery container still
+     only had room for one comfortably-wide card. Container queries
+     resolve against the nearest ancestor with `container-type`, which is
+     why the declaration must live on the wrapper, not the grid itself. */
+  .equal-columns {
+    container-type: inline-size;
+    container-name: integration-card-grid;
+  }
+
   /* Force equal columns in the card grid */
   .equal-columns :global(.card-grid) {
     display: grid;
     /* Default to a single column so cards keep a usable width and the EC
-       snippet doesn't overflow. The 2-column breakpoint is driven by a
-       container query below — the gallery container's own width, not the
-       viewport, is what matters because the Starlight sidebar consumes a
-       big chunk of viewport at desktop. */
+       snippet doesn't overflow. */
     grid-template-columns: 1fr;
     width: 100%;
     gap: 1rem;
-    container-type: inline-size;
-    container-name: integration-card-grid;
   }
 
   /* Ensure each card takes full space in its column */
@@ -224,10 +231,24 @@ const sortedIntegrations = visibleIntegrations.sort((a, b) => {
 
   /* Cap at two columns at all wider widths — never three. The breakpoint is
      2 × ~22rem (a comfortable card width that keeps the title readable and
-     the `aspire add` snippet from horizontally scrolling) plus the 1rem gap. */
+     the `aspire add` snippet from horizontally scrolling) plus the 1rem gap.
+     Below this threshold a single full-width card is more readable than two
+     squished ones. */
   @container integration-card-grid (min-width: 45rem) {
     .equal-columns :global(.card-grid) {
       grid-template-columns: repeat(2, minmax(0, 1fr));
+    }
+  }
+
+  /* Defensive fallback for the (vanishingly small) set of older browsers
+     that don't yet support container queries: use a viewport media query
+     tuned for desktop where the Starlight sidebar is visible. Modern
+     browsers ignore this because @container above wins on cascade. */
+  @supports not (container-type: inline-size) {
+    @media (min-width: 80rem) {
+      .equal-columns :global(.card-grid) {
+        grid-template-columns: repeat(2, minmax(0, 1fr));
+      }
     }
   }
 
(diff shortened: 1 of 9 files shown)


src/frontend/src/components/Integrations.astro
```diff
@@ -30,7 +30,17 @@ interface AvailableDoc {
 
 type AvailableDocs = AvailableDoc[];
 
-const sortedIntegrations = integrations.sort((a, b) => {
+// Feature flag — when 'true', show client integrations and the hosting/client
+// toggle group. Default (anything else) hides client cards from the gallery
+// page and removes the toggle group. The underlying data files, API docs, and
+// content articles continue to be generated/updated regardless of this flag.
+const showClientIntegrations = import.meta.env.PUBLIC_SHOW_CLIENT_INTEGRATIONS === 'true';
```
eerhardt, line 37: When would we ever use/change this feature flag?

src/frontend/src/components/Integrations.astro
```diff
@@ -30,7 +30,17 @@ interface AvailableDoc {
...
+
+const visibleIntegrations = showClientIntegrations
+  ? integrations
+  : integrations.filter((integration) => !(integration.tags ?? []).includes('client'));
```
eerhardt, line 41: Why don't we want to show client integrations?

Would it make more sense to have a filter at the top? To toggle hosting / client?

src/frontend/src/components/Integrations.astro
IEvangelist: If we convince ourselves that we want to show Client integrations on the integration's gallery. It felt wrong to just delete them. So, I decided to just hide them behind a flag for now.

src/frontend/src/components/Integrations.astro
IEvangelist: This is literally the behavior @davidfowl wanted me to remove. He wants client integrations gone.

davidfowl: From the top level search yeah. They will appear in "how you connenct to the hosting integration" instructions.

eerhardt (approved):

PR #1057, opened 2026-05-23
Update TypeScript AppHost GA docs

sebastienros:
TypeScript AppHosts are GA in the current docs, so the public guidance needs to describe the supported setup path and generated SDK location consistently.

## Summary

- Expands `aspire add` and `aspire restore` CLI docs to explain how TypeScript AppHosts update `aspire.config.json` and regenerate the SDK.
- Keeps the current docs aligned with the updated `apphost.mts` and `./.aspire/modules/aspire.mjs` generated SDK path from `release/13.4`.
- Clarifies brownfield TypeScript AppHost setup with supported AppHost-root package managers while preserving the existing Node.js 22.x/24.x LTS prerequisite wording.
- Updates FAQ links away from old TypeScript preview release-note anchors and keeps old release notes unchanged.
- Updates Twoslash virtual module wiring so current docs and historical snippets both compile.

Related to microsoft/aspire#17043.

## Validation

- `pnpm --dir ./src/frontend run test:unit:docs`
- `pnpm --dir ./src/frontend run test:unit:twoslash-blocks`
- `NODE_OPTIONS=--max-old-space-size=8192 pnpm --dir ./src/frontend run build:skip-search`


src/frontend/src/content/docs/reference/cli/includes/option-project.md
```diff
@@ -4,4 +4,4 @@ title: Option AppHost
 
 **`--apphost <apphost>`**
 
-The path to the Aspire AppHost project file.
+The path to the Aspire AppHost file or project file, such as `apphost.ts`, `apphost.cs`, or an AppHost `.csproj`.
```
mitchdenny: Minor: the example lists `apphost.ts`, but everywhere else in this PR (and the TypeScript AppHost docs) the canonical filename is `apphost.mts`. Suggest aligning to `apphost.mts` so users don't get a different extension here than `aspire init` actually produces.

```suggestion
The path to the Aspire AppHost file or project file, such as `apphost.mts`, `apphost.cs`, or an AppHost `.csproj`.
```

src/frontend/src/content/docs/ja/get-started/faq.mdx
```diff
@@ -63,13 +63,13 @@ Aspire は、実システムで頻出する多くの構成要素に対する統
 
 **いいえ。** .NET は Aspire で特に強力な体験を提供しますが、アプリ モデルは 1 つのフレームワークやランタイムに限定されません。Aspire は複数スタックにまたがるサービス調整を目的に設計されており、TypeScript の AppHost によって既存の JavaScript / TypeScript アプリにも自然に適合します。
 
-詳しくはこちら: [TypeScript AppHost サポート](/ja/whats-new/aspire-13-2/#-typescript-apphost-のサポート-プレビュー)
+詳しくはこちら: [TypeScript AppHost サポート](/ja/app-host/typescript-apphost/)
```
mitchdenny: The link target was updated from the 13.2 preview "what's new" anchor to the GA project-structure page, but the visible label still says "TypeScript AppHost サポート" (Support). The English FAQ was updated to "TypeScript AppHost project structure" — consider updating the Japanese label to match, e.g. `[TypeScript AppHost プロジェクト構成](/ja/app-host/typescript-apphost/)`. Same on line 72.

src/frontend/config/twoslash.config.mjs
```diff
@@ -85,7 +85,13 @@ export function readAspireTypes() {
  */
 export function getTwoslashExtraFiles() {
   const { source } = readAspireTypes();
-  return source ? { '.aspire/modules/aspire.mts': source } : {};
+  return source
+    ? {
+        '.aspire/modules/aspire.mts': source,
+        // Keep historical snippets compiling if they import the old generated path.
+        '.modules/aspire.ts': source,
```
mitchdenny: The comment frames this alias as backward-compat for "historical snippets," but the only place in the repo that still imports from `./.modules/aspire.js` is `src/frontend/src/components/AppHostBuilder.astro` — the live interactive AppHost builder, with ~30 snippet variants. Everywhere else (this PR's `typescript-apphost.mdx`, `add-aspire-existing-app.mdx`, `SimpleAppHostCode.shared.ts`, `data/twoslash/aspire.d.ts`, the doc-writer skill) uses the GA path `./.aspire/modules/aspire.mjs`.

Net effect: users who copy a snippet out of the interactive builder get an import path that doesn't match what `aspire init` actually generates, so the file won't resolve in their project.

Recommendation: migrate `AppHostBuilder.astro` to emit `./.aspire/modules/aspire.mjs` across all snippet variants and drop this alias. If that's out of scope for this PR, please reword the comment to make the temporary nature explicit, e.g. `// Temporary: AppHostBuilder.astro still imports from the pre-GA path; remove this alias once it's migrated to .aspire/modules/aspire.mjs.`

PR #1085, opened 2026-05-27
Block rendered twoslash failures

IEvangelist:
## Summary
- make the two-slash audit fail on any rendered TypeScript diagnostic instead of allowing known failures
- fix the current two-slash failures by improving the generated Aspire type surface and updating the stale GitHub Models sample
- add a reusable twoslash-validator skill and mention it in Copilot instructions

Prevents stuff rendering like this with errors:

https://aspire.dev/integrations/cloud/azure/azure-ai-foundry/azure-ai-foundry-host/#configure-a-foundry-project

image

## Validation
- pnpm test:unit:twoslash-blocks
- pnpm test:unit:twoslash-types
- pnpm test:unit:llms-txt
- pnpm lint


mitchdenny (approved):
Approving but I'm wondering if this is best implemented as a skill vs. some kind of static check?

IEvangelist:
> Approving but I'm wondering if this is best implemented as a skill vs. some kind of static check?

It's not just a skill. There's also new unit and e2e testing that validates we break the build to prevent ourselves from ever failing the UX live like this again. But the skill itself will be useful for ensuring AI doesn't mess things up.

PR #1112, opened 2026-05-28
Update docs for TypeScript AppHost GA

sebastienros:
TypeScript AppHosts are generally available in Aspire 13.4, so ASPIREATS001 should no longer be presented as an active experimental warning. This updates the docs to reflect the retired diagnostic and removes suppression guidance that would send users down the wrong path.

## Changes

- Retire ASPIREATS001 docs and remove it from diagnostics navigation and the diagnostics overview, while keeping the page available for users who still encounter it on older versions.
- Remove preview and ASPIREATS001 suppression guidance from multi-language integration authoring docs and scope current support to TypeScript AppHosts.
- Align the Japanese multi-language integration page and architecture metadata with the 13.4 TypeScript AppHost GA story.

## Validation

- `pnpm --dir ./src/frontend exec vitest run --config vitest.config.ts tests/unit/seo-lengths.vitest.test.ts`
- `pnpm --dir ./src/frontend run build:skip-search`


davidfowl (approved):

davidfowl:
Did you do a search for "preview" as well?

PR #1195, opened 2026-06-02
chore: omit empty TypeScript API modules

IEvangelist:
## Summary
- omit generated TypeScript API module JSON when a package has no functions or types
- remove existing empty TS module files, including the Community Toolkit Dapr entries shown in the API reference
- document the omission behavior in the generator docs and update-integrations skill

## Validation
- Not run locally; CI should validate this change.


alistairmatthews (approved):
Good to go.

PR #1200, opened 2026-06-03
Fix Mailpit environment variable names

patrickklaeren:
Variable names seem to have changed.

Docs reference `SMTP` prefix but this does not seem to be the case for 13.4

image


patrickklaeren:
@microsoft-github-policy-service agree

IEvangelist (approved):

PR #1204, opened 2026-06-03
Fix TypeScript documentation gaps

sebastienros:
## Summary
- Add supported TypeScript AppHost examples for docs-only gaps identified by the language audit
- Add TypeScript package/config guidance for Community Toolkit integrations
- Keep C#-only notes where TypeScript APIs are still unavailable

## Validation
- pnpm --dir ./src/frontend test:unit:twoslash-blocks
- git --no-pager diff --check

### DIFF
--- a/src/frontend/src/content/docs/integrations/observability/seq/seq-host.mdx
+++ b/src/frontend/src/content/docs/integrations/observability/seq/seq-host.mdx
@@ -163,7 +163,8 @@ import { ContainerLifetime, createBuilder } from './.aspire/modules/aspire.mjs';
 
 const builder = await createBuilder();
 
-const seq = await builder.addSeq("seq", "admin-password");
+const adminPassword = await builder.addParameter("seq-admin-password", { secret: true });
+const seq = await builder.addSeq("seq", adminPassword);
 await seq.withDataVolume();
 await seq.excludeFromManifest();
 await seq.withLifetime(ContainerLifetime.Persistent);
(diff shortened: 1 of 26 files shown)


src/frontend/src/content/docs/integrations/observability/seq/seq-host.mdx
```diff
@@ -158,7 +158,24 @@ var myService = builder.AddProject<Projects.ExampleProject>()
 </TabItem>
 <TabItem id="typescript" label="TypeScript">
 
-The TypeScript AppHost doesn't currently expose a `withDataVolume(...)` API for Seq. To persist Seq data from a TypeScript AppHost, use [`withBindMount`](/get-started/app-host/) to map a host directory into the container at `/data` instead.
+```typescript title="apphost.mts"
+import { ContainerLifetime, createBuilder } from './.aspire/modules/aspire.mjs';
+
+const builder = await createBuilder();
+
+const seq = await builder.addSeq("seq", "admin-password");
```
adamint: Could we avoid hardcoding the Seq admin password here too? The earlier TypeScript sample on this page uses `addParameter(..., { secret: true })`, which seems like the right pattern for copy/paste docs. This can use the same `seq-admin-password` parameter instead of putting the password literal in source.

adamint (commented):
found one issue

src/frontend/src/content/docs/integrations/observability/seq/seq-host.mdx
sebastienros: Good call. I updated the TypeScript data volume sample to use the secret `seq-admin-password` parameter instead of a literal password.

davidfowl (approved):

PR #1310, opened 2026-06-29
Align sidebar-page content with the search box (expanded and collapsed)

IEvangelist:
## Problem

On pages that render the topics sidebar (topic-nav pages like `/community/`, the get-started TOC pages, and the C#/TS API-reference pages), the main content (the page `H1` and all body content) was pushed to the **right** of the header search box's inline-start edge. The content column's left edge should line up with the search box's left edge.

It was **worse when the sidebar was collapsed** into its rail: the content shifted even further right instead of staying aligned.

## Root cause

`src/frontend/src/styles/site.css` reserved a fixed "shoulder" of inline-start padding on `.content-panel > .sl-container` to keep the floating collapse/expand toggle (`left: var(--sl-sidebar-width)`) from overlapping the `H1`:

- `padding-inline-start: 3.5rem` in the expanded state, and
- a separate override of `padding-inline-start: 5rem` in the collapsed/rail state.

But `.sl-container` is **right-aligned** within its panel (`margin-inline-start: auto`), so its inline-start edge already lines up with the search box. The fixed shoulder simply offset the content to the right of that aligned edge at every width, and the larger collapsed override made the rail state worse.

## Fix

Replace both the fixed expanded shoulder and the collapsed override with a single self-adjusting value that reserves toggle clearance **only** when the column's natural inline-start gutter doesn't already provide it, collapsing to `0` (flush with the search box) once it does:

```css
padding-inline-start: max(
  0px,
  calc(
    var(--aspire-sidebar-toggle-width) + 0.5rem - var(--sl-content-pad-x) -
      max(0px, calc(100% - var(--sl-content-width)))
  )
);
```

Because the toggle and the content panel share the sidebar's right edge, the sidebar width cancels out of the math, so the one expression is correct for **every sidebar width** (topic-nav, C#/TS API-ref) **and for both the expanded and collapsed/rail states** — when collapsed, the rail is narrower, the panel is wider, the gutter grows, and the shoulder collapses to `0` on its own. The now-obsolete collapsed override is removed.

## Validation

Measured on the live site via Playwright (CSS injected to match the committed rule), across `/community/` (non-TOC topic-nav), `/get-started/first-app/` (TOC), and `/reference/api/csharp/` (40rem API-ref sidebar), in **both** expanded and collapsed states, at 1280 / 1600 / 2000 / 2400px:

- **Wide viewports:** `H1` left == content-box left == search-box left (flush) in every page type and both sidebar states.
- **Narrow viewports:** the shoulder appears only when the content box would otherwise sit under the toggle, keeping the `H1` clear of it.
- Collapsed @2000px on the TOC page went from `H1` at 507px (80px right of search at 427px) to flush at 427px.

No local `pnpm build` was run (validated against production CSS via injection).

### DIFF
--- a/src/frontend/src/styles/site.css
+++ b/src/frontend/src/styles/site.css
@@ -1215,16 +1215,6 @@ li a.external-link::after {
     gap: 0;
   }
 
-  /* Drop the floating toggle's overlap-shoulder when collapsed — the rail
-     leaves enough room on its own (and the toggle still floats off the
-     rail's right edge thanks to left: var(--sl-sidebar-width)). */
-  :root[data-sidebar-collapsed]:has(.topics-sidebar[data-api-ref]) .content-panel > .sl-container,
-  :root[data-topic-sidebar-collapsed]:has(.topics-sidebar[data-topic-nav])
-    .content-panel
-    > .sl-container {
-    padding-inline-start: 5rem;
-  }
-
   /* Subtle blur-in for the filter row and sidebar entries as the rail
      expands. JS sets `data-sidebar-expanding` / `data-topic-sidebar-expanding`
      on <html> for ~320ms (just past the 250ms width transition), so this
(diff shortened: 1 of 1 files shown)


davidfowl:
UX changes need screenshots

davidfowl (approved):

PR #1323, opened 2026-07-01
Update Aspire VS Code dashboard docs

adamint:
The Aspire Dashboard no longer opens by default when starting an AppHost from VS Code, so update the extension docs to point users at the Open Dashboard CodeLens/command instead.

Also removes the old dashboard section and trims the gutter decoration details now that the screenshot covers the visual state.

Tested with `pnpm --dir src/frontend test:unit:docs`.


IEvangelist (approved):
LGTM, but please consider the automated PR feedback.

PR #1351, opened 2026-07-10
Replace Integration Data Updater agentic workflow with deterministic scripts

IEvangelist:
## Summary

Replaces the `gh aw` **Integration Data Updater** agentic workflow with a deterministic script + a thin, standard GitHub Actions workflow. Every step the agent performed was already deterministic, so this removes Copilot inference cost and the 20‑minute agent‑step cap that truncated API‑reference regeneration on catch‑up runs (see [run 29047474199](https://github.com/microsoft/aspire.dev/actions/runs/29047474199), which timed out at 20m after the auth fix in #1342 let it run).

## What changed

- **New:** `src/frontend/scripts/update-integration-data.ps1` — an OS/CI‑aware PowerShell orchestrator that:
  1. runs `pnpm update:all`,
  2. detects integration package **version** changes (compares committed `aspire-integrations.json` vs. the freshly generated one by `title` → `version`; metadata‑only changes do **not** trigger regen),
  3. conditionally regenerates the C# API JSON (`generate-package-json.ps1`), the TypeScript API JSON (`pnpm update:ts-api`), and the chained twoslash `aspire.d.ts` bundle,
  4. verifies the working tree only contains allowed data paths,
  5. captures icon‑resolution warnings and a file‑count summary, and emits the PR title/body + `GITHUB_OUTPUT`.
  Reproducible locally: `pwsh src/frontend/scripts/update-integration-data.ps1` (add `-SkipRegen` for a fast data‑only run).
- **New:** `.github/workflows/update-integration-data.yml` — schedule + `workflow_dispatch`, `timeout-minutes: 45`. Sets up .NET / Aspire CLI / pnpm / Node, runs the orchestrator, and — only when data changed — commits **only** the allowed generated files and opens a PR through the Aspire bot GitHub App, superseding prior open automation PRs. Mirrors the existing `update-release-branch.yml` app‑token pattern.
- **Removed:** `.github/workflows/update-integration-data.md` and its generated `.lock.yml`.
- **Updated:** the `update-integrations` skill automation note to describe the deterministic flow.

## Behavior notes

- Hard‑fails (red check, no PR) on: `update:all` failure, TS API regen failure, or any out‑of‑scope diff — so problems surface as CI rather than being silently "handled."
- Per‑package `pkgs/` generation failures (common for meta‑packages) are tolerated and reported in the PR body, matching prior behavior.
- Allowed data paths only: `aspire-integrations.json`, `github-stats.json`, `samples.json`, `assets/samples/**`, `data/pkgs/**`, `data/ts-modules/**`, `data/twoslash/aspire.d.ts`. Never workflow files, manifests, source, or docs (`integration-docs.json` reconciliation stays with the manual skill).

## Validation

- PowerShell AST parse: clean (pwsh 7.6).
- Workflow YAML: parses; 12 steps as expected.
- Version‑map parse verified against the real `aspire-integrations.json` (164/164 entries mapped); allowed‑path matcher unit‑checked (correctly excludes `integration-docs.json`, `package.json`, docs).
- Not run end‑to‑end locally (network + would mutate data); recommend a `workflow_dispatch` on this branch to smoke‑test in CI before merge.


src/frontend/scripts/update-integration-data.ps1
```diff
@@ -0,0 +1,391 @@
...
+function Write-Section([string]$Text) {
+    Write-Host ""
+    Write-Host "══ $Text ══" -ForegroundColor Cyan
+}
+
+function Invoke-Git {
+    param([Parameter(Mandatory)][string[]]$Arguments)
+    $out = & git -C $RepoRoot @Arguments 2>&1
+    return , @($out)
+}
```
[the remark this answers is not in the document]
IEvangelist: @
Good catch — refactored `Invoke-Git` in 9cd7790e to emit git output as flat strings straight to the pipeline (one line per object) and to normalize stderr `ErrorRecord`s to their message text, dropping the `return , @()` idiom entirely. Callers now reliably get `[string]` values, so `-join` and `Where-Object` behave, and `\0` is still the git exit code afterward. Verified against the `git show HEAD:` JSON-reconstruction and `status --porcelain | Where-Object` call sites.
@

src/frontend/scripts/update-integration-data.ps1
```diff
@@ -0,0 +1,391 @@
...
+    return [PSCustomObject]@{ Added = $added; Modified = $modified; Removed = $removed }
+}
+
+$pkgsCounts = Get-AreaCounts -Prefix 'src/frontend/src/data/pkgs/'
+$tsModulesCounts = Get-AreaCounts -Prefix 'src/frontend/src/data/ts-modules/'
+$twoslashChanged = [bool]((Invoke-Git @('diff', '--name-only', 'HEAD', '--', 'src/frontend/src/data/twoslash/aspire.d.ts')) |
+    Where-Object { $_ -and $_.Trim().Length -gt 0 })
+
+# ── PR title + body ─────────────────────────────────────────────────────────
+$dateShort = (Get-Date).ToString('M/d/yy')
```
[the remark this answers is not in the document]
IEvangelist: @
Fixed in 9cd7790e. `\` now uses `[DateTime]::UtcNow.ToString('M/d/yy', [CultureInfo]::InvariantCulture)`, so the PR title date is UTC (matching the `date -u` branch name and the UTC cron schedule) and the `/` separator renders consistently regardless of the runner's locale.
@

src/frontend/scripts/update-integration-data.ps1
```diff
@@ -0,0 +1,391 @@
...
+$anyChanges = [bool]$statusLines
+
+if (-not $anyChanges) {
+    Write-Section 'Result — no changes'
+    Write-Host "Integration data is already up to date." -ForegroundColor Green
+    Set-Output 'changed' 'false'
+    Set-Output 'versions_changed' 'false'
+    Set-Output 'regen_ran' 'false'
+    exit 0
+}
```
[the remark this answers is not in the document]
IEvangelist: @
Fixed in 9cd7790e. The change gate now runs `git status --porcelain` over the **full** working tree (no path filter) instead of only the two in-scope data dirs. So if `pnpm run update:all` touches anything at all, we fall through to the Phase 4 scope check, and an out-of-scope-only change fails loudly there rather than silently exiting 0.
@

src/frontend/scripts/update-integration-data.ps1
```diff
@@ -0,0 +1,407 @@
...
+        # Distinguish phase-2 vs phase-3 failure using the script's log markers.
+        if ($tsLog -match 'Twoslash type generation failed') {
+            Write-Error "Twoslash bundle generation failed. Aborting; no PR will be opened."
+        }
+        else {
+            Write-Error "TypeScript API generation failed. Aborting; no PR will be opened."
+        }
+        exit 1
+    }
+    $tsApiSummary = 'succeeded'
```
radical: [automated] On a **partial** TS-API regen failure this opens a PR that claims success. `generate-ts-api-json.ps1` counts per-package `aspire sdk dump`/transform failures into `$failed` and prints `Complete: N failed`, but it exits 0 (no `exit`/`throw` at the end), so `update-ts-api.ts` only fails on a *catastrophic* throw — not on individual package failures. So `$tsExit` is 0 here and we hardcode `$tsApiSummary = 'succeeded'`, then open a PR with stale/missing `ts-modules/*` and a twoslash bundle rebuilt without them (hover tooltips fall back to `any`), with nothing red.

Two problems with that:
- It contradicts the "Hard-fails (red check, no PR) on … TS API regen failure" note in the PR description — that's only true for total failure.
- It's a reporting regression vs the old body, which had a `TS API JSON: S succeeded, F failed` line, so a reviewer could see the count.

Suggest parsing the `Complete: X succeeded, Y failed` line the same way the C# path parses `Done! Success:`, surfacing `Y` in the body, and deciding whether `Y > 0` should fail the run or just be reported.

.github/workflows/update-integration-data.yml
```diff
@@ -0,0 +1,201 @@
...
+            exit 0
+          fi
+
+          # Reset the branch to a single fresh commit on top of the current main
+          # and force-push, so the open PR (if any) updates in place.
+          git checkout -B "$branch"
+          git commit \
+            -m "$PR_TITLE" \
+            -m "Co-authored-by: Copilot App <223556219+Copilot@users.noreply.github.com>"
+          git push --force origin "HEAD:refs/heads/$branch"
```
radical: [automated] Force-pushing the shared `chore/update-integration-data` branch every run will clobber any human commits on the open automation PR. The old model opened a fresh PR per run, so that couldn't happen.

The intended safety net looks like the `:octocat: auto-merge` label, but I don't see any workflow in `.github/workflows/` that consumes it — so same-day auto-merge isn't established in-repo, and a maintainer fix pushed onto the PR could be silently overwritten by the next scheduled run.

Worth confirming auto-merge is actually wired up (branch ruleset / external bot), or guarding the force-push when the branch has non-bot commits on top.

src/frontend/scripts/update-integration-data.ps1
```diff
@@ -0,0 +1,407 @@
...
+
+function Get-VersionMap {
+    param([Parameter(Mandatory)][string]$JsonText)
+    $map = @{}
+    if ([string]::IsNullOrWhiteSpace($JsonText)) { return $map }
+    $entries = $JsonText | ConvertFrom-Json
+    foreach ($entry in @($entries)) {
+        if ($entry.PSObject.Properties.Name -contains 'title' -and
+            $entry.PSObject.Properties.Name -contains 'version') {
+            $map[[string]$entry.title] = [string]$entry.version
```
radical: [automated] Non-blocking: this keys the version map on `title`, so two entries sharing a title would collapse (last wins) and a version move could be missed. It's fine today (164/164 entries, no duplicate titles) — just a latent assumption worth a dedup key or an assert.

Unrelated rollout reminder (not about this line): the reuse/stray-close logic in the workflow only matches the `chore/update-integration-data` branch family, so if any pre-existing agentic automation PR is open at merge time, close it by hand. None is open right now, so no action needed today.

radical (commented):
[automated] Automated review pass (findings validated with a second model). Three inline notes: one Medium (TS-API partial-failure reporting), one Medium (force-push clobber on the rolling branch), one non-blocking (title-keyed version map + rollout reminder).

src/frontend/scripts/update-integration-data.ps1
IEvangelist: Updated the orchestrator to parse the generator completion summary and abort when the summary is missing or any package failed. Successful runs now include the exact succeeded/failed/skipped counts in the PR body.

.github/workflows/update-integration-data.yml
IEvangelist: Added a guard that rejects replacement when the rolling branch contains any unmerged commit not authored and committed by the bot. The push now also uses an exact `--force-with-lease` to protect concurrent updates.

src/frontend/scripts/update-integration-data.ps1
IEvangelist: Added an explicit duplicate-title assertion in `Get-VersionMap`, so ambiguous integration data now aborts instead of silently collapsing entries.

IEvangelist:
Thank you @radical, this should be good to go now.

JamesNK (approved):
Looks good! Well-structured deterministic replacement for the agentic workflow. Clean phase-based orchestration, good scope safety with the allowed-path check, and the rolling branch pattern avoids stale PRs. No blocking issues found.

PR #1359, opened 2026-07-13
Post inline PR suggestions for forbidden words

IEvangelist:
## Summary

Extends the **Forbidden Words** check so that, in addition to failing a PR when a forbidden phrase is found, it now posts **inline PR review `suggestion` blocks** that rewrite each offending line using the rule''s replacement. Suggestions are posted by the **Aspire bot GitHub App** (the same app used elsewhere in this repo), and are skipped when a comment already exists on that line, so re-runs don''t pile up duplicates.

The blocking gate is preserved: the check still fails when forbidden words are present.

## How it works

A secure two-workflow split (so it works for fork PRs, which can''t see secrets on `pull_request`):

1. **`Forbidden Words`** (`pull_request`) — scans added lines, writes a `findings.json` + `meta.json`, uploads them as an artifact, and fails the check when violations exist. Read-only token, no secrets.
2. **`Forbidden Words Suggestions`** (`workflow_run`) — runs in the privileged base-repo context, mints the Aspire bot token, and posts the inline suggestions. It only reads the artifact JSON as data and runs **trusted scripts from the default branch** — it never checks out or executes PR code.

> [!NOTE]
> Because the suggestion workflow runs trusted scripts from the default branch, inline suggestions only start appearing on PRs **after this change merges to `main`**.

## Changes

- **`.github/forbidden-words.json`** — every rule now requires a `replacement`; updated `$comment` docs; excluded the new posting script.
- **`.github/scripts/check-forbidden-words.sh`** — validates replacements (fails if any rule lacks one), aggregates all matches on a line into a single corrected line, and writes a findings document. The replacement is inserted **verbatim** (matching still honors `caseSensitive`, but replacement text is never case-adjusted and does not support backreferences).
- **`.github/scripts/post-forbidden-word-suggestions.sh`** (new) — turns findings into a single batched PR review of `suggestion` blocks, deduped against existing comments, with a per-comment fallback.
- **`.github/workflows/forbidden-words.yml`** — scan (continue-on-error) → upload findings + PR metadata → gate/fail on violations.
- **`.github/workflows/forbidden-words-suggest.yml`** (new) — `workflow_run` follow-up that posts the suggestions.

## Validation

- Both scripts `bash -n` clean; both workflows pass `actionlint`.
- Scan logic tested locally against crafted diffs: detection, single aggregated suggestion per line, missing-`replacement` failure (exit 2), and verbatim replacement (incl. `$`/`@`/`\` edge cases).
- `pnpm build` not run (no frontend changes).

### DIFF
--- a/.github/scripts/check-forbidden-words.sh
+++ b/.github/scripts/check-forbidden-words.sh
@@ -6,11 +6,22 @@
 # pre-existing text is never flagged. Rules are defined in
 # .github/forbidden-words.json.
 #
+# In addition to failing when a forbidden phrase is found, this script records
+# a machine-readable list of findings (see FINDINGS_FILE below). Each finding
+# includes the corrected line produced by substituting every matched rule's
+# 'pattern' with its 'replacement'. A companion workflow
+# (.github/workflows/forbidden-words-suggest.yml) turns those findings into
+# inline pull request review suggestions.
+#
 # Usage:
 #   check-forbidden-words.sh <base_sha> <head_sha>
 #
 # Environment overrides:
-#   CONFIG_FILE  Path to the rules file (default: .github/forbidden-words.json)
+#   CONFIG_FILE    Path to the rules file (default: .github/forbidden-words.json)
+#   FINDINGS_FILE  When set, a JSON document of findings is written here (even
+#                  when violations are present). Shape:
+#                    { "findings": [ { "file", "line", "original",
+#                                      "suggestion", "messages", "patterns" } ] }
 #
 # Exits non-zero when at least one forbidden phrase is found.
 
@@ -20,6 +31,7 @@ BASE_SHA="${1:?base SHA required}"
 HEAD_SHA="${2:?head SHA required}"
 MERGE_BASE="$(git merge-base "$BASE_SHA" "$HEAD_SHA")" || { echo "::error::Unable to compute merge-base of $BASE_SHA and $HEAD_SHA"; exit 2; }
 CONFIG_FILE="${CONFIG_FILE:-.github/forbidden-words.json}"
+FINDINGS_FILE="${FINDINGS_FILE:-}"
 
 if [[ ! -f "$CONFIG_FILE" ]]; then
   echo "::error::Config file not found: $CONFIG_FILE"
@@ -27,15 +39,44 @@ if [[ ! -f "$CONFIG_FILE" ]]; then
 fi
 
 command -v jq >/dev/null 2>&1 || { echo "::error::jq is required"; exit 2; }
+command -v perl >/dev/null 2>&1 || { echo "::error::perl is required"; exit 2; }
 
 # Read global exclude globs (newline separated).
 mapfile -t GLOBAL_EXCLUDES < <(jq -r '.excludePaths // [] | .[]' "$CONFIG_FILE")
 
 rule_count=$(jq '(.rules // []) | length' "$CONFIG_FILE")
+
+# Emit an empty findings file up front so callers always have a valid document.
+write_findings() {
+  [[ -z "$FINDINGS_FILE" ]] && return 0
+  if [[ -s "$FINDINGS_TMP" ]]; then
+    jq -s '{findings: .}' "$FINDINGS_TMP" > "$FINDINGS_FILE"
+  else
+    printf '{"findings":[]}\n' > "$FINDINGS_FILE"
+  fi
+}
+
+FINDINGS_TMP="$(mktemp)"
+trap 'rm -f "$FINDINGS_TMP"' EXIT
+
 if [[ "$rule_count" -eq 0 ]]; then
   echo "No rules defined in $CONFIG_FILE; nothing to check."
+  write_findings
   exit 0
 fi
+
+# Every rule must define a non-empty replacement so a suggestion can be built.
+missing_replacement=$(jq '[.rules[] | select((.replacement // "") == "")] | length' "$CONFIG_FILE")
+if [[ "$missing_replacement" -ne 0 ]]; then
+  echo "::error::Every rule in $CONFIG_FILE must define a non-empty \"replacement\". Found $missing_replacement rule(s) without one."
+  exit 2
+fi
+
+# Preload rule fields to avoid re-invoking jq for every line.
+mapfile -t PATTERNS       < <(jq -r '.rules[].pattern' "$CONFIG_FILE")
+mapfile -t REPLACEMENTS   < <(jq -r '.rules[].replacement' "$CONFIG_FILE")
+mapfile -t MESSAGES       < <(jq -r '.rules[] | (.message // "")' "$CONFIG_FILE")
+mapfile -t CASE_SENSITIVE < <(jq -r '.rules[] | (.caseSensitive // false)' "$CONFIG_FILE")
 
 # Returns 0 if $1 matches any of the shell globs passed as remaining args.
 matches_any_glob() {
@@ -51,6 +92,20 @@ matches_any_glob() {
   return 1
 }
 
+# Substitutes every occurrence of a pattern with its replacement, inserting the
+# replacement verbatim. Perl inserts the value of $r without reprocessing its
+# contents, so no backreferences ($1) or escape sequences are interpreted.
+# Matching honors the rule's case sensitivity.
+apply_replacement() {
+  CONTENT="$1" PAT="$2" REP="$3" CS="$4" perl -e '
+    my $c = $ENV{CONTENT};
+    my $p = $ENV{PAT};
+    my $r = $ENV{REP};
+    if ($ENV{CS} eq "true") { $c =~ s/$p/$r/g; } else { $c =~ s/$p/$r/gi; }
+    print $c;
+  '
+}
+
 violations=0
 
 # List the files changed between base and head (added/copied/modified/renamed).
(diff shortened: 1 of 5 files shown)


.github/workflows/forbidden-words-suggest.yml
```diff
@@ -0,0 +1,114 @@
...
+    types: [ completed ]
+
+permissions:
+  contents: read
+  actions: read
+  pull-requests: write
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.workflow_run.id }}
+  cancel-in-progress: false
```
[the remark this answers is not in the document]
IEvangelist: Good catch. Fixed in 3cedc084 — concurrency is now keyed by the source PR (`head_repository.full_name` + `head_branch`) with `cancel-in-progress: true`, so rapid pushes for the same PR cannot run concurrently and race the dedupe check. Using the head repo name keeps same-named branches on different forks in separate groups.

.github/scripts/post-forbidden-word-suggestions.sh
```diff
@@ -0,0 +1,117 @@
...
+if [[ ! -f "$FINDINGS_FILE" ]]; then
+  echo "Findings file not found ($FINDINGS_FILE); nothing to suggest."
+  exit 0
+fi
+
+total=$(jq '.findings | length' "$FINDINGS_FILE")
+if [[ "$total" -eq 0 ]]; then
+  echo "No findings; nothing to suggest."
+  exit 0
+fi
```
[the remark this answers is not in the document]
IEvangelist: Fixed in 3cedc084 — the findings file is now treated as untrusted input: the script verifies it is a valid JSON document with a `.findings` array and skips quietly with a warning instead of failing when it is missing or invalid.

.github/scripts/post-forbidden-word-suggestions.sh
```diff
@@ -0,0 +1,117 @@
...
+
+# Existing review comments on the PR, as {path, line} objects (line falls back to
+# original_line for outdated comments). Used to avoid duplicate suggestions.
+existing_json=$(gh api --paginate "/repos/$REPO/pulls/$PR_NUMBER/comments" \
+  --jq '.[] | {path: .path, line: (.line // .original_line)}' | jq -s '.')
+
```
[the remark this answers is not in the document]
IEvangelist: Fixed in 3cedc084 — listing existing comments is now guarded; on any `gh api` failure the script logs a warning and exits 0 (skips) rather than failing the job, so a transient / permissions / rate-limit error will not break this best-effort workflow (and it avoids posting duplicates).

.github/scripts/check-forbidden-words.sh
```diff
@@ -20,23 +31,53 @@ BASE_SHA="${1:?base SHA required}"
...
+  write_findings
   exit 0
 fi
 
+# Every rule must define a non-empty replacement so a suggestion can be built.
+missing_replacement=$(jq '[.rules[] | select((.replacement // "") == "")] | length' "$CONFIG_FILE")
+if [[ "$missing_replacement" -ne 0 ]]; then
+  echo "::error::Every rule in $CONFIG_FILE must define a non-empty \"replacement\". Found $missing_replacement rule(s) without one."
+  exit 2
+fi
```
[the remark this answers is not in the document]
IEvangelist: Fixed in 3cedc084 — the missing-`replacement` path now calls `write_findings` (emitting an empty findings document) before `exit 2`, so callers that always consume the findings artifact still get a valid document while the check still fails.

.github/workflows/forbidden-words-suggest.yml
```diff
@@ -0,0 +1,117 @@
...
+  # forks with the same branch name in separate groups.
+  group: ${{ github.workflow }}-${{ github.event.workflow_run.head_repository.full_name }}-${{ github.event.workflow_run.head_branch }}
+  cancel-in-progress: true
+
+jobs:
+  suggest:
+    name: Post inline suggestions
+    runs-on: ubuntu-latest
+    # Only act on runs triggered by pull requests.
+    if: ${{ github.event.workflow_run.event == 'pull_request' }}
```
joperezr, line 33: How are we guaranteed to know we can trust the meta.json from the checkout? From what I understand here, we are not locking to the head SHA, so theoretically a fork PR could make changes to the meta.json and make bad suggestions to a _different_ PR.

Before we merge this in, we should make sure we have thought well through this and ensure that we have mitigations to protect against this.

.github/scripts/check-forbidden-words.sh
```diff
@@ -91,41 +160,68 @@ for file in "${CHANGED_FILES[@]}"; do
...
+        violations=$((violations + 1))
+        line_messages+=("$message")
+        line_patterns+=("$pattern")
+
         printf '::error file=%s,line=%s::Forbidden phrase (/%s/): %s\n' \
           "$file" "$lineno" "$pattern" "$message"
         printf '    %s:%s: %s\n' "$file" "$lineno" "$content"
-        violations=$((violations + 1))
+
+        corrected="$(apply_replacement "$corrected" "$pattern" "$replacement" "$case_sensitive")"
```
joperezr: NIT: Accrording to the agent, this may backfire if you have a pattern defined which uses special PCRE2 notation that Perl doesn't support (like `\p{Xan}`) it will crash the run and make it so no suggestions are posted ever. I know this is unlikely since we are mostly just using simple regex and mostly forbidden words, but there may be a cheap way to fix that broken case if you just do something like `corrected="$(apply_replacement ... || printf '%s' "$corrected")"` which would no longer crash the full run any more.

IEvangelist: Handled this by catching replacement-engine failures per line. An unsupported pattern now emits a warning and omits only that line's suggestion, allowing the scan and all other valid suggestions to continue.

.github/workflows/forbidden-words-suggest.yml
IEvangelist: Removed the artifact-provided PR coordinates entirely. The privileged workflow now matches the trusted `workflow_run` head SHA, repository, and branch against GitHub's open pull requests and proceeds only when exactly one PR matches.

.github/scripts/post-forbidden-word-suggestions.sh
```diff
@@ -0,0 +1,132 @@
...
+  --argjson existing "$existing_json" \
+  '
+  $f[0].findings
+  | map(select(. as $find
+      | ($existing | any(.path == $find.file and .line == $find.line)) | not))
+  | map({
+      path: .file,
+      line: .line,
+      side: "RIGHT",
+      body: ((.messages | join("\n")) + "\n\n```suggestion\n" + .suggestion + "\n```")
```
joperezr: One thing that might still be worth hardening: `messages` comes straight out of the artifact, which is built by the PR's *own* copy of the scan script + `forbidden-words.json`. So a fork author fully controls it, and we post it verbatim as the Aspire bot — meaning they can slip `@mentions` in (bot pings arbitrary people) or drop fake "approved by maintainers"-style text into the comment body.

The coordinates are now safely re-resolved from the workflow_run identity, which is great — it's just the body content that's still trusted. Could we re-derive the message text from the trusted `forbidden-words.json` on `main` (keyed off `patterns`, which we already carry in the finding) instead of echoing the artifact's `messages`?

.github/scripts/check-forbidden-words.sh
```diff
@@ -91,41 +160,74 @@ for file in "${CHANGED_FILES[@]}"; do
...
-      continue
-    fi
-
-    grep_opts=(-P)
-    if [[ "$case_sensitive" != "true" ]]; then
-      grep_opts+=(-i)
-    fi
+  # Check each added line against every applicable rule, aggregating all matches
+  # on the line into a single corrected line (at most one suggestion per line).
+  while IFS= read -r entry; do
```
joperezr, line 165: Something the agent found that I'm not 100% but worth double checking:

>Not on this exact line — it's about the awk that builds `$added_lines` just above (the `/^\+\+\+/ { next }` rule). With `--unified=0`, an added line whose *content* starts with `++` shows up as `+++...` in the diff, so that rule skips it as if it were the `+++ b/file` header.
Two knock-on effects: the forbidden phrase on that line never gets flagged (gate silently passes), and since `line++` is skipped too, every later added line in the hunk ends up off-by-one — so otherwise-valid suggestions land on the wrong line or get rejected (422). Matching the header form only should fix it, e.g. `/^\+\+\+ / { next }` (note the trailing space).

joperezr (commented):
Two more comments but this is almost ready.

.github/scripts/post-forbidden-word-suggestions.sh
IEvangelist: The poster now ignores artifact-provided messages and maps each finding's patterns to messages from the trusted default-branch `forbidden-words.json`. It also rejects unknown, malformed, multiline, or fence-breaking findings so untrusted artifact content cannot escape the suggestion block.

.github/scripts/check-forbidden-words.sh
IEvangelist: Confirmed and fixed. The parser now processes `+` lines only after entering a diff hunk, which distinguishes file headers from added content without misclassifying lines that begin with `++` or shifting subsequent line numbers.

IEvangelist:
Thank you @joperezr, all set.

joperezr (approved):
LGTM

PR #1368, opened 2026-07-16
Fix Open dropdown to show brand icons instead of generic sparkle

IEvangelist:
## Problem

Every item in the page **Open** dropdown ("Open in GitHub Copilot", "Open in Claude", "Open in ChatGPT") rendered the same generic "AI" sparkle glyph instead of each service's brand icon.

## Root cause

The `starlight-page-actions` plugin (v0.6.2) renders a real brand icon for each **built-in** action, but hardcodes the generic `AIIcon` sparkle for every **custom** action. Our config defined ChatGPT, Claude, and GitHub Copilot as `custom` actions, so all three fell back to the sparkle.

### Before

image

## Fix

- **`astro.config.mjs`** — Replace the `custom` block with the plugin's built-in `chatgpt` / `claude` / `githubCopilot` actions. The built-in hrefs and labels are identical to what we had, so behavior is unchanged, and ChatGPT and Claude now show their real brand marks.
- **`components/starlight/PageTitle.astro`** — The plugin's built-in "GitHub Copilot" glyph is the Copilot mascot rather than the GitHub logo. Override it with the site's existing GitHub octocat (Starlight's `github` icon path) via a masked-icon rule scoped to `[data-action-icon='githubCopilot']`, matching the GitHub logo used elsewhere on the site.

### After

image

## Validation

- `node --check astro.config.mjs` passes.
- GitHub octocat data-URI verified as well-formed SVG.
- No tests assert on the dropdown labels/config.
- `pnpm build` intentionally not run locally — please verify the rendered dropdown on CI/preview.

### DIFF
--- a/src/frontend/astro.config.mjs
+++ b/src/frontend/astro.config.mjs
@@ -92,23 +92,16 @@ export default defineConfig({
         plugins: [
           starlightPageActions({
             share: true,
+            // Use the plugin's built-in actions so each entry renders its real
+            // brand icon. Custom actions are always given the plugin's generic
+            // "AI" sparkle glyph, which is why every "Open in …" item previously
+            // showed the same icon. The built-in hrefs match what we want, and
+            // the GitHub Copilot mark is restyled to the site's GitHub logo in
+            // src/components/starlight/PageTitle.astro.
             actions: {
-              chatgpt: false,
-              claude: false,
-              custom: {
-                copilot: {
-                  label: 'Open in GitHub Copilot',
-                  href: 'https://github.com/copilot/?prompt=',
-                },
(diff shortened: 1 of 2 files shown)


JamesNK (approved):
Looks good! The approach is correct — using built-in actions instead of custom ones is the right fix to get brand icons.

One minor nit: in `PageTitle.astro`, the second CSS rule block (the `:global(#dropdown-menu .dropdown-item .action-icon[data-action-icon='githubCopilot'])` rule) is missing 2-space indentation — it starts at column 0 while all other rules in that `<style>` block are indented. Non-blocking but worth fixing for consistency.

PR #1370, opened 2026-07-17
Fix topic sidebar toggle styling

IEvangelist:
## Summary
- keep topic landing-page collapse and expand icons transparent and borderless when no On this page bar is present
- preserve the existing bar styling on API reference and topic pages that render the mobile TOC

## Validation
- verified the landing page and regular topic page at affected viewport widths using the local dev server and playwright-cli


JamesNK (approved):
Looks good — clean split of positioning vs. visual styling using ID selectors, and the comment update accurately reflects the new behavior.

PR #1378, opened 2026-07-17
Correct Community Toolkit add-on and migration docs

IEvangelist:
> [!IMPORTANT]
> This PR is ready for review, but it should be rebased and merged after #1377 because it performs the series' final package-mapping and sidebar reconciliation.

## Summary

- Correct the Community Toolkit add-on guidance for Redis, MongoDB, MySQL, PostgreSQL, SQL Server, Keycloak, JavaScript monorepos, and MassTransit RabbitMQ against the actual decorator and configuration APIs.
- Correct Bun, Go, and Python graduation guidance, including removal of the deprecated Python extensions package and the source-backed TypeScript exception for virtual-environment opt-out.
- Model the complete Flyway ordering chain, explain the generated TypeScript helper's ordering limitation, and add runnable synchronized C#/TypeScript examples for exported Bun and Keycloak operations.
- Reconcile package destinations and sidebar labels, and preserve complete hero artwork.

## Validation

- Verified APIs and behavior against `CommunityToolkit/Aspire` commit `b9d40479805b74ea20b7dc57cb659d46e6f11c43`.
- External Toolkit AppHost, generated TypeScript bindings, MassTransit client, and official Bun/Go/Python scratch samples compiled or type-checked successfully.
- Focused page metadata, SEO, twoslash, structured-data, integration-update, and locale/sidebar suites pass; mapping assertions and affected browser routes pass.
- Combined landing rehearsal passes 8 focused suites/153 tests, 132 unique mappings, 248 logo renders, and 199 synchronized AppHost tab groups with zero browser failures or console errors.

No local `pnpm build` was run.

PR #1376, opened 2026-07-17
Split Dapr, DAB, and LavinMQ integration docs

IEvangelist:
> [!IMPORTANT]
> This PR is ready for review, but it should be rebased and merged after #1372, #1375, and #1374 because the series shares mapping, sidebar, and logo behavior.

## Summary

- Replace the standalone Dapr, Data API Builder, and LavinMQ pages with nine focused get-started, host, and connect pages, including legacy redirects, sidebar entries, and package mappings.
- Document the source-verified hosting and connection contracts: Dapr endpoint variables, DAB service-discovery values, and LavinMQ's exact RabbitMQ-compatible connection properties and fixed default host ports.
- Provide complete synchronized C# and TypeScript AppHost examples, plus C#, TypeScript, Python, and Go consuming-client examples where stable clients exist.
- Preserve complete logo artwork, including theme-aware Dapr assets.

## Validation

- Verified APIs, tests, generated baselines, `[AspireExport]` metadata, and TypeScript modules against `CommunityToolkit/Aspire` commit `b9d40479805b74ea20b7dc57cb659d46e6f11c43`.
- Fresh external C# and TypeScript AppHosts restored and compiled/type-checked; C#, TypeScript, Python, and Go client samples passed their native checks.
- Focused redirects, sidebar, mapping, SEO, and twoslash tests pass; all nine routes and three legacy redirects pass browser checks.
- Combined landing rehearsal passes 8 focused suites/153 tests, 248 logo renders, and 199 synchronized AppHost tab groups with zero browser failures or console errors.

No local `pnpm build` was run.

PR #1375, opened 2026-07-17
Audit data and AI Community Toolkit integration docs

IEvangelist:
> [!IMPORTANT]
> This PR is ready for review, but it should be rebased and merged after #1372 because the series relies on the shared mapping and contained-logo guardrails.

## Summary

- Audit all 18 Ollama, KurrentDB, Meilisearch, RavenDB, SQLite, and SurrealDB get-started, host, and connect pages against `CommunityToolkit/Aspire` commit `b9d40479805b74ea20b7dc57cb659d46e6f11c43`.
- Correct runtime, connection, client, and resource-chain guidance, and replace C#-only capability callouts with complete synchronized C# and TypeScript AppHost examples.
- Make every new AppHost example runnable, use language-neutral framing, and preserve the precise source-backed exceptions for APIs that are not exported.
- Keep every integration logo contained and use theme-specific KurrentDB assets so the complete artwork remains visible in light and dark themes.

## Validation

- Verified TypeScript API shapes against the pinned Toolkit source, `[AspireExport]` metadata, and checked-in TypeScript AppHost fixtures.
- Focused page metadata, SEO, and twoslash suites pass: 3 files, 65 tests.
- Combined landing rehearsal passes 8 focused suites, 153 tests, `check-data`, mapping uniqueness, and `git diff --check`.
- Browser audit confirms all 18 routes render and their logos remain contained and legible at desktop/mobile widths in light/dark themes; synchronized AppHost tabs render without console errors.

No local `pnpm build` was run.

PR #1386, opened 2026-07-21
Fast-follow: address non-blocking maintainer feedback from #1372-#1378

IEvangelist:
## Fast-follow: non-blocking maintainer feedback from #1372–#1378

This PR closes the deferred, non-blocking review nits left on the merged community integration PRs. Each item is either implemented or explicitly closed with current, source-backed evidence. The diff is surgical (17 files) and preserves runnable C#/TypeScript AppHost parity, package mappings, redirect compatibility, and logo legibility.

### 1. ThemeImage containment without invalid rendered markup (review on #1372)

- **Change (test-only):** Added focused render coverage in `custom-components.vitest.test.ts` asserting a non-square ThemeImage is contained via the `getImage(...)` optimizer — the emitted `data-light`/`data-dark` optimizer URLs carry `fit=contain`, while the rendered `` does **not** carry a `fit="contain"` attribute.
- **Evidence:** `ThemeImage.astro` is byte-for-byte unchanged from `upstream/main` (the fix already landed with #1372: `fit: 'contain'` lives at the `getImage` optimizer-options level and is not spread onto `<Image>` props). The CSS `object-fit: contain` behavior is untouched. The new test codifies the intended containment contract without asserting invalid HTML.
- **Visual verification:** The KurrentDB wordmark asset is **3420×755** (extremely wide, non-square) requested at 100×100. Without `fit: contain` at the optimizer level it would distort/crop; with it, the full wordmark is preserved (reinforced by CSS `object-fit: contain`). Because `ThemeImage.astro` is unchanged from what already ships, rendering is identical to production; the containment contract is now additionally guarded by the automated render test rather than a manual browser pass.

### 2. Mapping destination helper coverage (review on #1372)

- **Change:** Made `docsPageExists` injectable and extended it so `.md`/`index.md` resolve in addition to `.mdx`/`index.mdx`, preventing future valid Markdown mappings from false-negativing. Added `describe('docsPageExists markdown resolution')` regression coverage (flat/index × `.md`/`.mdx`, a false case, and query/hash stripping) in `update-integrations.vitest.test.ts`.

### 3. Deferred dev-tool accuracy pass (review on #1374)

Every claim was source-verified against the latest Aspire CLI and current CommunityToolkit/Aspire source:

- **`aspire add flagd` → `aspire add communitytoolkit-flagd`** and **`aspire add mailpit` → `aspire add communitytoolkit-mailpit`** (corrected). The Aspire CLI derives the friendly name via `IntegrationPackageSearchService.GenerateFriendlyName`: `packageId.Replace("Aspire.Hosting.", "").Replace('.', '-').ToLowerInvariant()`. For `CommunityToolkit.Aspire.Hosting.Flagd` / `...MailPit` this yields exactly `communitytoolkit-flagd` / `communitytoolkit-mailpit`. This also matches the existing `communitytoolkit-powershell` / `communitytoolkit-rust` precedent in the repo.
- **C# `AddOfrepProvider()` → `AddOfrepProvider(options => { /* leave BaseUrl unset → OFREP_ENDPOINT */ })`** (corrected). `OpenFeature.Providers.Ofrep`'s `FeatureBuilderExtensions` exposes only two overloads — `AddOfrepProvider(Action<OfrepProviderOptions>)` and `AddOfrepProvider(string domain, Action<OfrepProviderOptions>)`. There is **no** parameterless overload, so the previous `AddOfrepProvider();` would not compile. The source's `CreateProvider` confirms that when `BaseUrl` is unset it falls back via `OfrepOptions.FromConfiguration(...)` to the `OFREP_ENDPOINT` env var injected by the AppHost, so the added comment is accurate.
- **`Aspire:GoFeatureFlag:Client` config binding — retained as correct (no change).** `CommunityToolkit.Aspire.GoFeatureFlag`'s `AspireGoFeatureFlagExtensions` defines `private const string DefaultConfigSectionName = "Aspire:GoFeatureFlag:Client";`, so the documented key is accurate.

### 4. Twoslash coverage for data/AI TypeScript examples (review on #1375)

- **Change:** Converted the runnable TypeScript AppHost examples (fences titled `apphost.mts`) on all six pages added by #1375 — Ollama, KurrentDB, Meilisearch, RavenDB, SQLite, SurrealDB — from plain fences to canonical `twoslash` fences so CI type-checks them. (The `index.ts` consuming-app client blocks are out of scope — they are not AppHost examples.)
- **Evidence:** The generated `src/frontend/src/data/twoslash/aspire.d.ts` already exposes all six builder methods (`addOllama`, `addKurrentDB`, `addMeilisearch`, `addRavenDB`, `addSqlite`, `addSurrealServer`), so every converted block compiles with **zero** diagnostics. No suppressions/allowlists were added (per the twoslash-validator skill).

### 5. Dapr Azure Container Apps caution (review on #1376) — no change (obsolete)

- **Decision:** The old caution ("Azure Container Apps does not retain Dapr settings on redeploy") is **obsolete** for the current declarative workflow and is intentionally **not** reintroduced.
- **Evidence:** The original caution applied to the old manual `PublishAsAzureContainerApp(...)` approach, where Dapr enabled outside IaC could be lost on redeploy. The section now uses declarative `environment.WithDaprComponents()` (CommunityToolkit `AzureContainerAppEnvironmentResourceBuilderExtensions`), which emits `ContainerAppDaprConfiguration { IsEnabled = true, ... }` into the ACA Bicep and re-emits every `IDaprComponentResource` into the generated infrastructure on **every** `aspire publish`. Because Dapr is re-asserted from source on each deploy, the redeploy-loss caveat no longer applies. Current (2025) ACA declarative IaC deployments retain Dapr settings, corroborating the removal.

### 6. Canonical internal links after host-only splits (review on #1377)

- **Change:** Repointed in-repo links that still hit legacy host-only redirect routes to their canonical `*-get-started` pages, across all docs content (not just the reviewer's examples): `community/thanks.mdx` (Deno/Java/PowerShell/Rust), `get-started/app-host.mdx` (Java), `languages-and-runtimes/index.mdx` (Java/Rust/PowerShell), and the `ja` equivalents (pointing at the canonical non-`ja` English get-started targets, matching the existing `ja` redirect targets and the established `go/` precedent). Redirects in `config/redirects.mjs` are left intact for external compatibility.
- **Scope guard:** JavaScript/Python/Orleans/MAUI targets are real standalone pages and were left untouched; Go was an earlier split and is already canonical everywhere.
- **Test:** Added a focused regression test in `redirects.vitest.test.ts` that scans all docs for links to the legacy #1377 host-only routes and fails if any reappear, while the existing test still verifies the redirects themselves are preserved.

### 7. Confirm already-landed #1378 prose nits — confirmed, no change

- `golden path` no longer appears anywhere; `go-host.mdx` and `python.mdx` say **"recommended approach."**
- `postgresql-extensions.mdx` states **"`migrations.WaitFor(database)` ensures Flyway starts only after PostgreSQL is ready."**
- Both nits are already present on latest `main`; no duplicate edits were made.

## Validation

- `pnpm test:unit:twoslash-blocks` — passed (0 diagnostics; all converted blocks type-check).
- `pnpm test:unit:contracts` (includes redirects) — passed, including the new legacy-route regression test.
- `pnpm test:unit:components` and the update-integrations structured-data tests — passed, including the new ThemeImage containment and `docsPageExists` tests.
- `git diff --check` — clean.
- Source-backed sample validation for every changed/claimed API (Aspire CLI friendly-name algorithm; OpenFeature Ofrep overloads; CommunityToolkit GoFeatureFlag config section; CommunityToolkit Azure Dapr IaC emission).
- Per repo convention, `pnpm build` was **not** run locally.

Based on `upstream/main` at `c35a18d1bae710ec7d3ed255aa307d6f1cc53765`.


src/frontend/tests/unit/redirects.vitest.test.ts
```diff
@@ -88,3 +89,63 @@ test('host-only integration routes redirect to their get-started pages', () => {
...
+function normalizeLinkPath(url: string): string {
+  return url
+    .split('#')[0]
+    .split('?')[0]
+    .replace(/\/+$/, '');
+}
+
+test('docs do not link to legacy host-only routes (use canonical get-started URLs)', () => {
+  // Matches markdown links `](/path)` and JSX `href="/path"` / `href='/path'`.
+  const linkPattern = /(?:\]\(|href=["'])(\/[^)"'\s#?]*)/g;
```
joperezr, line 131: Tiny non-blocking nit: this regex only catches `](/…)` and `href="/…"` links. Reference-style defs (`[x]: /route`) and JSX `href={expr}` would slip past it. Nothing in the docs hits those forms today so the guard works fine as-is but flagging just in case since the agent said this might be an issue in the future.

joperezr (approved):
Thanks David, LGTM

PR #1397, opened 2026-07-23
fix: Normalize terminology in sample updates

IEvangelist:
## Summary

- centralize Aspire terminology normalization for generated data
- normalize sample titles, descriptions, README content, and AppHost code before writing `samples.json`
- prevent future generated catalogs from reintroducing `.NET Aspire` or `app host`

This fixes the sample-update regression visible in #1393.

## Tests

- `pnpm --dir ./src/frontend exec vitest run --config vitest.config.ts tests/unit/update-samples.vitest.test.ts tests/unit/update-integrations.vitest.test.ts`
- replayed the `samples.json` content from #1393 through the normalizer (8 deprecated occurrences reduced to 0)

### DIFF
--- a/src/frontend/scripts/aspire-terminology.ts
+++ b/src/frontend/scripts/aspire-terminology.ts
@@ -1,0 +1,8 @@
+export function normalizeAspireTerminology(text: string): string {
+  return text
+    .replace(/\b([Aa]) \.NET Aspire\b/gi, (_match, article: string) =>
+      article === 'A' ? 'An Aspire' : 'an Aspire'
+    )
+    .replace(/\.NET Aspire/gi, 'Aspire')
+    .replace(/\bapp host\b/gi, 'AppHost');
+}
--- a/src/frontend/scripts/aspire-terminology.ts
+++ b/src/frontend/scripts/aspire-terminology.ts
@@ -1,8 +1,19 @@
+const horizontalWhitespace = String.raw`[ \t]+`;
+const legacyAspirePattern = String.raw`\.NET${horizontalWhitespace}Aspire`;
+const legacyAppHostPattern = String.raw`\bapp${horizontalWhitespace}host\b`;
+
+const aspireWithArticle = new RegExp(
+  String.raw`\b([Aa])${horizontalWhitespace}${legacyAspirePattern}\b`,
+  'gi'
+);
+const legacyAspire = new RegExp(legacyAspirePattern, 'gi');
+const legacyAppHost = new RegExp(legacyAppHostPattern, 'gi');
+
 export function normalizeAspireTerminology(text: string): string {
   return text
-    .replace(/\b([Aa]) \.NET Aspire\b/gi, (_match, article: string) =>
+    .replace(aspireWithArticle, (_match, article: string) =>
       article === 'A' ? 'An Aspire' : 'an Aspire'
     )
-    .replace(/\.NET Aspire/gi, 'Aspire')
-    .replace(/\bapp host\b/gi, 'AppHost');
+    .replace(legacyAspire, 'Aspire')
+    .replace(legacyAppHost, 'AppHost');
 }
--- a/src/frontend/scripts/aspire-terminology.ts
+++ b/src/frontend/scripts/aspire-terminology.ts
@@ -1,9 +1,16 @@
 const horizontalWhitespace = String.raw`[ \t]+`;
-const legacyAspirePattern = String.raw`\.NET${horizontalWhitespace}Aspire`;
+// Require `.NET` to sit at a non-word boundary so longer tokens such as
+// `ASP.NET Aspire` or `Microsoft.NET Aspire` are left intact instead of being
+// corrupted into `ASPAspire` / `MicrosoftAspire`.
+const legacyAspirePattern = String.raw`(?<!\w)\.NET${horizontalWhitespace}Aspire`;
 const legacyAppHostPattern = String.raw`\bapp${horizontalWhitespace}host\b`;
 
+// Markdown emphasis/link openers can sit between the article and the term in
+// raw README/AppHost content (e.g. `a **.NET Aspire**` or `a [.NET Aspire](url)`).
+// Consuming them keeps the article grammatical after the term is shortened.
+const markdownOpeners = String.raw`[*\[]*`;
 const aspireWithArticle = new RegExp(
-  String.raw`\b([Aa])${horizontalWhitespace}${legacyAspirePattern}\b`,
+  String.raw`\b([Aa])${horizontalWhitespace}(${markdownOpeners})${legacyAspirePattern}\b`,
   'gi'
 );
 const legacyAspire = new RegExp(legacyAspirePattern, 'gi');
@@ -11,8 +18,10 @@ const legacyAppHost = new RegExp(legacyAppHostPattern, 'gi');
 
 export function normalizeAspireTerminology(text: string): string {
   return text
-    .replace(aspireWithArticle, (_match, article: string) =>
-      article === 'A' ? 'An Aspire' : 'an Aspire'
+    .replace(
+      aspireWithArticle,
+      (_match, article: string, opener: string) =>
+        `${article === 'A' ? 'An' : 'an'} ${opener}Aspire`
     )
     .replace(legacyAspire, 'Aspire')
     .replace(legacyAppHost, 'AppHost');
--- a/src/frontend/scripts/aspire-terminology.ts
+++ b/src/frontend/scripts/aspire-terminology.ts
@@ -2,16 +2,24 @@ const horizontalWhitespace = String.raw`[ \t]+`;
 
 // A deprecated term only matches when it isn't fused to an adjacent alphanumeric
 // character, so longer tokens like `ASP.NET Aspire` or `.NET AspireX` stay intact
-// while `_`, `*`, backticks, and `[` still count as boundaries (Markdown emphasis,
-// inline code, and links wrap terms with those characters).
+// while punctuation and Markdown wrappers (`_`, `*`, `[`) still count as
+// boundaries around a standalone term.
 const termStart = String.raw`(?<![A-Za-z0-9])`;
 const termEnd = String.raw`(?![A-Za-z0-9])`;
 
-// Markdown emphasis/link/code openers can sit between an article and a term in
-// raw README/AppHost content (e.g. `a **.NET Aspire**`, `a [.NET Aspire](url)`, or
+// Markdown emphasis/link openers can sit between an article and a term in raw
+// README content (e.g. `a **.NET Aspire**`, `a [.NET Aspire](url)`, or
 // `a _.NET Aspire_`), so the article corrector consumes them to stay grammatical
-// after the term shrinks.
-const markdownOpeners = String.raw`[*\[_\x60]*`;
+// after the term shrinks. Backticks are intentionally excluded: inline code is
+// skipped wholesale (see `codeRegion`), so a term inside it is never rewritten.
+const markdownOpeners = String.raw`[*\[_]*`;
+
+// Fenced code blocks and inline code spans are copied through verbatim so sample
+// commands like `dotnet aspire run` are never rewritten into an unrunnable
+// `Aspire run`. Everything outside these regions is treated as prose. This runs
+// over raw README Markdown (`readmeRaw` feeds the Copy/View Markdown actions), so
+// preserving code exactly matters.
+const codeRegion = /```[\s\S]*?```|~~~[\s\S]*?~~~|`[^`\n]*`/g;
 
 interface TerminologyRule {
   /**
@@ -51,7 +59,36 @@ const terminologyRules: readonly TerminologyRule[] = [
   },
 ];
 
-export function normalizeAspireTerminology(text: string): string {
+export function normalizeAspireTerminology(text: string): string;
+export function normalizeAspireTerminology(
+  text: string | null | undefined
+): string | null | undefined;
+export function normalizeAspireTerminology(
+  text: string | null | undefined
+): string | null | undefined {
+  if (text == null) {
+    return text;
+  }
+
+  return normalizeProse(text);
+}
+
+// Apply every terminology rule to prose only, copying fenced and inline code
+// regions through untouched so sample commands stay runnable.
+function normalizeProse(text: string): string {
+  let result = '';
+  let lastIndex = 0;
+
+  for (const match of text.matchAll(codeRegion)) {
+    result += applyRules(text.slice(lastIndex, match.index));
+    result += match[0];
+    lastIndex = match.index + match[0].length;
+  }
+
+  return result + applyRules(text.slice(lastIndex));
+}
+
+function applyRules(text: string): string {
   return terminologyRules.reduce(applyRule, text);
 }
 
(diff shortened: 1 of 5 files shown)


src/frontend/scripts/aspire-terminology.ts
```diff
@@ -0,0 +1,19 @@
+const horizontalWhitespace = String.raw`[ \t]+`;
+const legacyAspirePattern = String.raw`\.NET${horizontalWhitespace}Aspire`;
```
JamesNK: This pattern also matches the `.NET Aspire` substring inside a longer term. For example, `ASP.NET Aspire application` becomes `ASPAspire application`, and `Microsoft.NET Aspire` becomes `MicrosoftAspire`. Because this helper runs over arbitrary README and AppHost source text, a future input can be corrupted rather than merely normalized. Please require `.NET` not to be preceded by a word character and add regression cases for these inputs.

src/frontend/scripts/aspire-terminology.ts
```diff
@@ -0,0 +1,19 @@
...
+const legacyAppHostPattern = String.raw`\bapp${horizontalWhitespace}host\b`;
+
+const aspireWithArticle = new RegExp(
```
JamesNK: The article correction only recognizes unformatted text, but `readme` and `readmeRaw` are raw Markdown. `A **.NET Aspire** project` currently becomes `A **Aspire** project`, and `A [.NET Aspire](https://aspire.dev/) project` becomes `A [Aspire](https://aspire.dev/) project`. That leaves the generated docs grammatically incorrect. Please handle Markdown wrappers/text nodes when correcting the article and cover bold/link cases in the tests.

JamesNK (changes requested):
## Review summary

**Phase A source checkpoint:** `microsoft/aspire` `main` at `e6388016c6c4b90f47589c15e9714975dc4aebc4` (2026-07-24). **Claims:** 0 product API/config/CLI claims were introduced, so there are no source verdicts to report.

**Phase B:** No documentation route or rendered example changed; 0 critical issues and 0 warnings. The docs server was not started because there was no affected page to exercise and the requested review excluded a website build.

The generation-code review found two blocking normalization cases, documented inline.

## Phase A - Claim verification

No changed file introduces a factual claim about Aspire APIs, runtime behavior, configuration, CLI surface, or package versions.

<details>
<summary>Verified and verified-with-nuance claims</summary>

None; the changed prose is implementation-local commentary rather than Aspire product documentation.

</details>

## Phase B - Doc-tester results

# Documentation Test Report

**Focus Area:** PR #1397 changed-file route mapping  
**Date:** 2026-07-24  
**Tester:** doc-tester agent  
**PR ref:** `ec1ec43481ddfce284d53cddf6c65859beb613c5`  
**Local server:** Not started - no documentation routes were affected.

## Summary

| Category | Passed | Failed | Warnings |
|----------|--------|--------|----------|
| Content Accuracy | 0 | 0 | 0 |
| Code Examples | 0 | 0 | 0 |
| CLI Commands | 0 | 0 | 0 |
| Links | 0 | 0 | 0 |

## Critical Issues

None.

## Warnings

None.

## Passed Checks

- Changed-file mapping confirmed that the PR only changes generator scripts, a sample utility comment, and unit tests; no rendered documentation page or route is modified.

## Recommendations

None for rendered documentation.

## Knowledge Gaps

None.

src/frontend/scripts/aspire-terminology.ts
IEvangelist: Good catch - fixed in 16cf407b. `legacyAspirePattern` now starts with a `(?<!\w)` lookbehind, so `.NET` only matches at a non-word boundary. `ASP.NET Aspire` and `Microsoft.NET Aspire` are now left intact instead of collapsing to `ASPAspire` / `MicrosoftAspire`, and I added regression cases for both.

src/frontend/scripts/aspire-terminology.ts
IEvangelist: Fixed in 16cf407b. The article corrector now consumes leading Markdown emphasis/link openers (`**`, `[`) between the article and the term, so `a **.NET Aspire**` becomes `an **Aspire**` and `a [.NET Aspire](https://aspire.dev/)` becomes `an [Aspire](https://aspire.dev/)`. Added bold and inline-link test cases to cover both.

src/frontend/tests/unit/update-samples.vitest.test.ts
```diff
@@ -0,0 +1,112 @@
...
+      appHostCode: null,
+    };
+
+    expect(normalizeSampleTerminology(sample)).toEqual(sample);
+  });
+
+  test('keeps generated sample data free of deprecated terminology', () => {
+    const textFields = ['title', 'description', 'readme', 'readmeRaw', 'appHostCode'] as const;
+    const deprecatedTerminology = new RegExp(
+      [legacyAspireName, legacyAppHostName]
```
JamesNK: This generated-data invariant omits `legacyDotnetAspireName`, even though both `terminologyRules` and `.github/forbidden-words.json` include that phrase. Please include all three configured terms so checked-in sample data cannot silently retain `dotnet aspire`.

src/frontend/scripts/aspire-terminology.ts
```diff
@@ -0,0 +1,67 @@
+const horizontalWhitespace = String.raw`[ \t]+`;
+// Markdown emphasis/link openers can sit between an article and a term in raw
+// README/AppHost content (e.g. `a **.NET Aspire**` or `a [.NET Aspire](url)`), so
+// the article corrector consumes them to stay grammatical after the term shrinks.
+const markdownOpeners = String.raw`[*\[]*`;
```
JamesNK: The Markdown article fix is still incomplete for valid `_` emphasis and inline-code wrappers. Because `_` and `` ` `` aren't in this opener class, `A _.NET Aspire_ project` and `A \`.NET Aspire\` project` normalize to `A _Aspire_ project` and `A \`Aspire\` project`, retaining the wrong article. This helper runs over raw README Markdown, so please handle these wrappers too and add regression cases.

src/frontend/scripts/aspire-terminology.ts
```diff
@@ -0,0 +1,67 @@
...
+   */
+  readonly article?: 'a' | 'an';
+}
+
+// Adding a rule here normalizes a new term everywhere generated sample and
+// integration data is written. Keep this list in sync with the deprecated terms
+// in `.github/forbidden-words.json`.
+const terminologyRules: readonly TerminologyRule[] = [
+  {
+    pattern: String.raw`(?<!\w)\.NET${horizontalWhitespace}Aspire`,
```
JamesNK: Please add the trailing `\b` required by the rule contract above. `applyRule` uses this pattern directly, so `.NET AspireX application` becomes `AspireX application`; with an article, `A .NET AspireX` becomes `A AspireX` because only the article pass adds an end boundary. This can corrupt a longer token in arbitrary README/AppHost text, the same class of issue as the now-fixed prefix case. Please add a suffix regression case too.

JamesNK (changes requested):
Static review of `5273ce9ae51636f9416d9bf976c0a1da7dd6d0ae` against PR base `0df9b83ee31925baf6ab98f8e07579020be8bba7`.

The three terminology rules match `.github/forbidden-words.json`, and the new call sites cover generated integration descriptions plus sample text fields. I found three remaining issues in the normalizer and its generated-data guard, noted inline.

No documentation route is changed, so there was no affected page to exercise. Per reviewer instruction, I did not run builds, tests, generators, or a development server.

src/frontend/scripts/aspire-terminology.ts
IEvangelist: Fixed in 30923d83. Rather than patch a trailing `\b` onto this one rule, I removed per-rule boundary responsibility entirely: every term core is now wrapped centrally by `boundedTerm` with alphanumeric edge guards (`(?<![A-Za-z0-9])` ... `(?![A-Za-z0-9])`). So `.NET AspireX` and `A .NET AspireX` are both left untouched, and no rule can silently forget a boundary again. Added suffixed-token regression cases (with and without a preceding article).

src/frontend/tests/unit/update-samples.vitest.test.ts
IEvangelist: Good call - fixed in 30923d83. The invariant now includes `legacyDotnetAspireName` alongside the other two, so checked-in sample data can't silently retain the `dotnet aspire` spelling.

src/frontend/scripts/aspire-terminology.ts
IEvangelist: Fixed in 30923d83. The old `(?<!\w)` guard treated `_` as a blocker, which is exactly why `_`-wrapped terms misbehaved. Switching to `(?<![A-Za-z0-9])` (and a matching `(?![A-Za-z0-9])` suffix) means `_` and backticks now count as boundaries, and I added them to `markdownOpeners` so the article correction fires. `A _.NET Aspire_ project` and `A `.NET Aspire` project` now normalize to `An _Aspire_ project` / `An `Aspire` project`. Added regression cases for both.

JamesNK (approved):
Reviewed the terminology normalization logic, generated-field coverage, boundary handling, and tests. No actionable issues found.

adamint (dismissed):
I don't think this is safe to merge as-is. The normalizer is applied to `readmeRaw` and `appHostCode` (`update-samples.ts:124,126`), but it's a plain regex pass over raw Markdown and C#, so it rewrites inside code fences, inline code, and string literals.

I ported `aspire-terminology.ts` at this head (`30923d83`) and ran it. Real output:

```
IN : "```bash\ndotnet aspire run\n```"
OUT: "```bash\nAspire run\n```"

IN : "Run `dotnet aspire run` to start."
OUT: "Run `Aspire run` to start."

IN : "var cmd = \"dotnet aspire run --project ./AppHost\";"
OUT: "var cmd = \"Aspire run --project ./AppHost\";"
```

`Aspire run` isn't a command anyone can paste. The `.NET Aspire` → `Aspire` rule substitutes a fixed capitalized string, so it can't produce the lowercase `aspire run` a shell would need. This matters because `readmeRaw` is what the Copy / View Markdown buttons serve (`src/utils/samples.ts:195-196`) and `appHostCode` renders through `<Code lang="csharp">` on the sample page — so we'd be handing people broken commands.

Two more from the same pass:

```
IN : "# A .NET Aspire sample\n\nSee [overview](#a-net-aspire-sample)."
OUT: "# An Aspire sample\n\nSee [overview](#a-net-aspire-sample)."
```
The heading gets rewritten, the anchor doesn't, so in-page links break.

```
IN : "The app.host file and my-app host"      OUT: "The app.host file and my-AppHost"
IN : ".NET ASPIRE and DOTNET ASPIRE"          OUT: "Aspire and Aspire"
```
The boundaries are alphanumeric-only, so `-` isn't a boundary, and case isn't preserved.

To be fair about severity: I checked the committed `samples.json` at this head and there are currently **zero** matches for `dotnet aspire`, `.NET Aspire`, or `app host`, so nothing is corrupted today. But this script exists to run against future upstream READMEs, and `dotnet aspire run` in a fenced block is about the most likely thing for one to contain. It'd land silently in generated data.

What I'd do: normalize the prose fields only (`title`, `description`, `readme` if that's rendered prose) and leave `appHostCode` alone entirely — it's C#, there's no prose to fix. For the Markdown fields, run it through an AST pass that skips `code`/`inlineCode` nodes, link destinations, and headings that already have inbound anchors. The repo already has remark in the dependency graph, so that's not a big lift.

Smaller things:
- `normalizeAspireTerminology(null)` throws (`TypeError`). `PackageRecord.description` is typed non-null and the call sites guard, but the previous optional-chaining behavior was more forgiving — worth a guard or an explicit contract.
- Worth adding tests for the cases above. The current tests don't cover fenced commands, inline code, anchors, case, or idempotence. Idempotence does hold, I checked.

Credit where it's due: the boundary work is solid — `ASP.NET Aspire` and `.NET AspireX` both stay intact, which is the part I expected to be broken and wasn't.

Also worth asking: `.github/forbidden-words.json` + `check-forbidden-words.sh` already gate this terminology in CI. This silently rewrites third-party README content instead of failing, which is a different policy than what we apply to our own docs. Is silent rewriting what we want for content we don't own?

IEvangelist:
Thanks for the careful review @adamint — especially porting the script and running it. The "un-pasteable command" case is exactly the failure mode I wanted to avoid, so I've made the normalizer code-safe in aba2cb5b.

### Fixed

- **Code fences, inline code, and C# are no longer rewritten.** The normalizer now copies fenced blocks (` ``` ` / `~~~`) and inline code spans (`` ` ``) through verbatim, applying rules only to the prose gaps between them. Your three cases now round-trip unchanged:
  - ` ```bash\ndotnet aspire run\n``` ` → identical
  - `` Run `dotnet aspire run` to start. `` → identical
  - `readmeRaw` prose is still normalized, but any embedded command survives.
- **`appHostCode` is no longer normalized at all.** It renders through `<Code lang="csharp">`, so there's no prose to fix and every reason not to touch identifiers, string literals, or comments. Removed from `normalizeSampleTerminology` (`update-samples.ts`) with a comment explaining why.
- **`normalizeAspireTerminology(null)` no longer throws.** It's now null-safe via overloads — `string → string`, and `null`/`undefined` pass through unchanged.
- **Tests added** for fenced commands, inline-code commands, a fenced legacy brand, mixed prose+code, idempotence (prose and mixed), and the null/undefined contract. The generated-data invariant is now idempotence-based (`normalize(x) === x` for `title`/`description`/`readme`/`readmeRaw`), which is both stronger than the old fixed-term scan and robust to code fences.

### On the AST suggestion

I went with string-region masking (skip fenced + inline code) rather than a full remark round-trip. It covers the realistic corruption vectors you found without adding a parser dependency, and — importantly for `readmeRaw`, which the Copy/View Markdown buttons serve byte-for-byte — it leaves the raw Markdown exactly as authored outside prose. Happy to revisit an AST pass if we later need to skip link destinations too, but I wanted to keep this "simple and clean."

### Intentional (with rationale)

- **Case:** `.NET ASPIRE` → `Aspire` is deliberate. `Aspire` is the canonical brand regardless of input casing, so normalizing to it (not preserving `ASPIRE`) is the intended outcome.
- **Hyphen boundary:** treating punctuation (`-`, `_`, `*`, `[`) as a boundary is by design — JamesNK specifically asked for `_`/`*`/`[` to count as boundaries. Making `-` a *non*-boundary on the start side would be asymmetric with the end side (`.NET Aspire-based` → `Aspire-based` relies on `-` being a boundary there). `my-app host` fusing is rare enough that I'd rather not risk that regression.
- **Heading anchors:** the on-page renderer regenerates heading slugs from the (normalized) text, so the auto-generated TOC stays consistent. The only breakage is a hand-authored `[text](#old-slug)` link pointing at the pre-normalization slug, which is narrow. A correct fix needs slug-aware in-doc link rewriting; I left it out to keep the change small, but can add it if you think it's worth it.

### On the policy question

Fair to raise. The way I see it, these are two different layers: `check-forbidden-words.sh` gates **first-party docs we author**, where failing the build is the right call. This normalizer runs at generation time over **third-party upstream READMEs** we surface but don't own — we can't fail someone else's README, so normalizing terminology for display is the pragmatic path (which is why the `ci:skip-forbidden-words` label is on this PR). That said, if the team would rather surface a warning/report than silently rewrite, I'm open to it — happy to discuss.

I checked the same thing you did: zero matches in `samples.json` today, so this is future-proofing rather than a fix for current corruption. Re-requesting your review — thanks again.

PR #1408, opened 2026-07-28
Revise interaction service API documentation

JamesNK:
Too much `IInteractionService`.

### DIFF
--- a/src/frontend/src/content/docs/extensibility/interaction-service.mdx
+++ b/src/frontend/src/content/docs/extensibility/interaction-service.mdx
@@ -28,11 +28,11 @@ If you need to collect input for a [custom resource command](/fundamentals/custo
 The `IInteractionService` API is not yet available in the TypeScript AppHost SDK.
 </Aside>
 
-## The `IInteractionService` API
+## The interaction service API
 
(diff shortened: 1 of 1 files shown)


adamint (approved):
Approving.

I checked the one thing that can actually break with a heading reword: the anchor slug. It does change — `the-iinteractionservice-api` → `the-interaction-service-api`, because the old heading had inline code in it — but nothing links to either fragment. Grepped this repo and `microsoft/aspire` for both slugs and for `interaction-service#`, and the only inbound references are page-level (`docs.topics.ts`, `aspireinteraction001.mdx:89`, `aspire-9-4.mdx:251`, and `tools/ReleaseNotes/data/whats-new-94.md` in the main repo), none with a fragment.

Built the site and served the page to confirm the rendered result: H2 renders as `The interaction service API` with id `the-interaction-service-api`, the "On this page" TOC points at the new slug, `#the-interaction-service-api` scrolls to the heading, and there were no console or page errors. `starlight-links-validator` reported "All internal links are valid" across all 13,857 pages, `test:unit` was 267/267, and lint was clean.

One thing worth knowing rather than fixing: external bookmarks to `…/interaction-service/#the-iinteractionservice-api` won't resolve after this. Fragments never reach the server, so `redirects.mjs` can't cover it. Not worth holding the PR for — I'd only bother with a compatibility anchor if we had evidence someone deep-links that heading, and I don't see any.

Reads better this way, and `IInteractionService` is still introduced two lines above the changed sentence, so the type name isn't lost.

PR #1410, opened 2026-07-28
Redact connection-string passwords in generated package docs

IEvangelist:
## Problem

The scheduled pipeline that mirrors public `main` to the internal AzDO remote (and merges it into `deploy` / `deploy-vnext-release`) is blocked by 1ES / CredScan **push protection**:

```
VS403654:BypassableBlock The push was rejected because it contains one or more secrets.
/src/frontend/src/data/pkgs/Aspire.Hosting.SqlServer.13.4.6.json(1032,116-124) : SEC101/037 : SqlLegacyCredentials
```

The flagged token is a **documentation placeholder**, not a real secret. `PackageJsonGenerator` copies package XML doc comments verbatim into the frontend data JSON, and several packages document an example connection string that embeds a literal `Password=password`:

- `Aspire.Hosting.SqlServer` - `Server=host,port;User ID=sa;Password=password;TrustServerCertificate=true`
- `Aspire.Hosting.PostgreSQL`, `CommunityToolkit.Aspire.Hosting.Minio`, `CommunityToolkit.Aspire.Hosting.SurrealDb`

`.config/CredScanSuppressions.json` already lists `password`, but that file is only honored by the Guardian **CI** CredScan task - the server-side push protection ignores it and scans per-commit.

## Fix

Sanitize the placeholder at the source so it stops recurring:

- New `DocumentationSanitizer.RedactConnectionStringPasswords`, applied to text / inline-code / code-block doc nodes **and** the `<example>` extraction paths in `CanonicalModelBuilder`. It rewrites connection-string `Password=`/`Pwd=` literals to `{password}`. The match requires no whitespace around `=`, so C# default parameter values (`string? password = null`) are untouched.
- `{password}` is used (rather than `<password>`) because it is **Markdown-safe** - angle brackets are dropped as raw HTML when doc nodes render to Markdown - and it matches the placeholder convention already used elsewhere in the generated data (e.g. `mysql://{user}:{password}@{host}`).
- Regenerated the four affected `pkgs/*.json` files (1 line each).
- Unit tests for the sanitizer (24 passing), including idempotency and a Markdown-safety assertion.

## Note on the one-time bypass

Push protection scans per-commit, so this forward fix cannot retroactively clean the value already committed in history. The currently-stuck sync still needs a **one-time soft-block bypass** in the AzDO Advanced Security portal (admin action). After that, this change keeps future data refreshes clean.

### DIFF
--- a/src/tools/PackageJsonGenerator/Helpers/CanonicalModelBuilder.cs
+++ b/src/tools/PackageJsonGenerator/Helpers/CanonicalModelBuilder.cs
@@ -81,7 +81,7 @@ internal sealed class CanonicalModelBuilder(Compilation compilation)
                 {
                     Name = f.Name,
                     Value = Convert.ToInt64(f.ConstantValue),
-                    Description = ExtractSummary(f),
+                    Description = DocumentationSanitizer.RedactConnectionStringPasswords(ExtractSummary(f)),
                 })];
         }
 
@@ -738,7 +738,7 @@ internal sealed class CanonicalModelBuilder(Compilation compilation)
                                         {
                                             Kind = "href",
                                             Value = href,
-                                            Text = child.Value.Trim() is { Length: > 0 } t ? t : null,
+                                            Text = child.Value.Trim() is { Length: > 0 } t ? DocumentationSanitizer.RedactConnectionStringPasswords(t) : null,
                                         });
                                     }
                                 }
@@ -840,7 +840,7 @@ internal sealed class CanonicalModelBuilder(Compilation compilation)
                                 {
                                     Kind = "href",
                                     Value = href,
-                                    Text = child.Value.Trim() is { Length: > 0 } t ? t : null,
+                                    Text = child.Value.Trim() is { Length: > 0 } t ? DocumentationSanitizer.RedactConnectionStringPasswords(t) : null,
                                 });
                             }
                             break;
--- a/src/tools/PackageJsonGenerator/Helpers/CanonicalModelBuilder.cs
+++ b/src/tools/PackageJsonGenerator/Helpers/CanonicalModelBuilder.cs
@@ -705,6 +705,7 @@ internal sealed class CanonicalModelBuilder(Compilation compilation)
                     // Normalize whitespace within text runs (newlines → spaces, collapse runs)
                     text = text.Replace("\r\n", " ").Replace("\n", " ").Replace("\r", " ");
                     while (text.Contains("  ")) text = text.Replace("  ", " ");
+                    text = DocumentationSanitizer.RedactConnectionStringPasswords(text);
                     if (!string.IsNullOrEmpty(text))
                     {
                         nodes.Add(new DocNode { Kind = "text", Text = text });
@@ -766,7 +767,7 @@ internal sealed class CanonicalModelBuilder(Compilation compilation)
                         }
 
                         case "c":
-                            nodes.Add(new DocNode { Kind = "code", Text = child.Value });
+                            nodes.Add(new DocNode { Kind = "code", Text = DocumentationSanitizer.RedactConnectionStringPasswords(child.Value) });
                             break;
 
                         case "code":
@@ -779,7 +780,7 @@ internal sealed class CanonicalModelBuilder(Compilation compilation)
                             nodes.Add(new DocNode
                             {
                                 Kind = "codeblock",
-                                Text = child.Value,
+                                Text = DocumentationSanitizer.RedactConnectionStringPasswords(child.Value),
                                 Language = lang,
                                 Value = region, // reuse value for region
                             });
--- a/src/tools/PackageJsonGenerator/Helpers/CanonicalModelBuilder.cs
+++ b/src/tools/PackageJsonGenerator/Helpers/CanonicalModelBuilder.cs
@@ -899,7 +899,9 @@ internal sealed class CanonicalModelBuilder(Compilation compilation)
         {
             // No code block — try to use the whole content as code
             var plainText = ExtractPlainText(exampleElement);
-            return plainText is not null ? new DocExample { Code = plainText } : null;
+            return plainText is not null
+                ? new DocExample { Code = DocumentationSanitizer.RedactConnectionStringPasswords(plainText)! }
+                : null;
         }
 
         var lang = NormalizeLanguage(
@@ -921,7 +923,7 @@ internal sealed class CanonicalModelBuilder(Compilation compilation)
                     while (text.Contains("  ")) text = text.Replace("  ", " ");
                     if (!string.IsNullOrWhiteSpace(text))
                     {
-                        descNodes.Add(new DocNode { Kind = "text", Text = text.Trim() });
+                        descNodes.Add(new DocNode { Kind = "text", Text = DocumentationSanitizer.RedactConnectionStringPasswords(text.Trim()) });
                     }
                     break;
 
@@ -942,7 +944,7 @@ internal sealed class CanonicalModelBuilder(Compilation compilation)
 
         return new DocExample
         {
-            Code = codeElement.Value,
+            Code = DocumentationSanitizer.RedactConnectionStringPasswords(codeElement.Value)!,
             Language = lang,
             Description = descNodes.Count > 0 ? descNodes : null,
             Region = region,
--- a/src/tools/PackageJsonGenerator/Helpers/DocumentationSanitizer.cs
+++ b/src/tools/PackageJsonGenerator/Helpers/DocumentationSanitizer.cs
@@ -1,0 +1,53 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Text.RegularExpressions;
+
+namespace PackageJsonGenerator.Helpers;
+
+/// <summary>
+/// Sanitizes documentation text copied verbatim from package XML docs before it
+/// is emitted into the generated JSON data files.
+/// </summary>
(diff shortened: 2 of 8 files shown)


src/tools/PackageJsonGenerator/Helpers/DocumentationSanitizer.cs
```diff
@@ -0,0 +1,53 @@
...
+
+namespace PackageJsonGenerator.Helpers;
+
+/// <summary>
+/// Sanitizes documentation text copied verbatim from package XML docs before it
+/// is emitted into the generated JSON data files.
+/// </summary>
+public static class DocumentationSanitizer
+{
+    private const string PasswordPlaceholder = "<password>";
```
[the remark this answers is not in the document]
IEvangelist: Good catch, fixed in c9749cd4. Switched the placeholder from `<password>` to `{password}` so it survives Markdown rendering. Turns out `{...}` is already the convention in this generated data (e.g. `mysql://{user}:{password}@{host}` in the MySql docs), so this also makes it consistent. Also excluded `{`/`}` from the regex value class to keep the redaction idempotent, and added a test asserting the placeholder contains no angle brackets.

src/tools/PackageJsonGenerator/Helpers/CanonicalModelBuilder.cs
```diff
@@ -705,6 +705,7 @@ private static List<DocNode> CollectNodes(XElement element)
                     // Normalize whitespace within text runs (newlines → spaces, collapse runs)
                     text = text.Replace("\r\n", " ").Replace("\n", " ").Replace("\r", " ");
                     while (text.Contains("  ")) text = text.Replace("  ", " ");
+                    text = DocumentationSanitizer.RedactConnectionStringPasswords(text);
                     if (!string.IsNullOrEmpty(text))
```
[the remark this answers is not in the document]
IEvangelist: Fixed in c9749cd4. `ExtractDocExample` now runs `RedactConnectionStringPasswords` on all three paths that previously bypassed it: the plain-text `Code` fallback, the description text nodes, and `Code = codeElement.Value`. So passwords inside `<example>` descriptions/code are redacted too and won't re-trigger CredScan.

src/tools/PackageJsonGenerator/Helpers/DocumentationSanitizer.cs
```diff
@@ -0,0 +1,56 @@
...
+
+namespace PackageJsonGenerator.Helpers;
+
+/// <summary>
+/// Sanitizes documentation text copied verbatim from package XML docs before it
+/// is emitted into the generated JSON data files.
+/// </summary>
+public static class DocumentationSanitizer
+{
+    private const string PasswordPlaceholder = "{password}";
```
radical: [automated] Recommend using `Placeholder` for the redaction token instead of `{password}` — this is the value 1ES recommends for scrubbed credential examples in generated content.

```csharp
private const string PasswordPlaceholder = "Placeholder";
```

Two things need to move alongside the const so the build stays green:

- The `{password}` expectations in `DocumentationSanitizerTests.cs`. In particular `RedactConnectionStringPasswords_IsIdempotent` leans on the braces being excluded from the value class (an already-redacted `{password}` can't re-match). With a bare `Placeholder` the sanitizer *will* re-match its own output — harmlessly, since re-running yields an identical string — but the test should assert `apply-twice == apply-once` rather than "the redacted value is left untouched".
- The 4 regenerated `pkgs/*.json` files that currently embed `{password}`.

src/tools/PackageJsonGenerator/Helpers/DocumentationSanitizer.cs
```diff
@@ -0,0 +1,56 @@
...
+    // Matches connection-string style "key=value" credential pairs such as
+    // "User ID=sa;Password=password" or "Pwd=hunter2". The '=' is intentionally
+    // required with no surrounding whitespace so that C# default parameter
+    // values ("string? password = null") are never matched. The value class
+    // stops at connection-string / JSON delimiters (';', quotes, backslash) and
+    // at the '{'/'}' (and legacy '<'/'>') placeholder markers, which keeps the
+    // replacement idempotent (an already-redacted "Password={password}" is left
+    // untouched).
+    private static readonly Regex ConnectionStringPasswordRegex = new(
+        "\\b(password|pwd)=[^;\"'{}<>\\s\\\\,]+",
```
radical: [automated] The value class here — `[^;"'{}<>\s\\,]+` — excludes `;`, quotes, braces, angle brackets, backslash, comma and whitespace, but not `` ` ``, `)`, `]`, `&` or `|`. So on those characters the match runs past the value and swallows the delimiter. Running the exact pattern + options through .NET 10:

```
Use `Password=secret` in config    ->  Use `Password={password} in config
See [docs](https://x?password=s)   ->  See [docs](https://x?password={password}
https://h/api?password=s&uid=sa    ->  https://h/api?password={password}
The default is Password=secret.    ->  The default is Password={password}
```

`csharp-api-markdown.ts:889` emits text nodes verbatim, so a dropped backtick / paren / bracket breaks inline-code pairing and link syntax in the generated Markdown. The 4 files regenerated in this PR only escape this because their values happen to be followed by `"` (which is excluded); the next package won't necessarily be.

Adding `` ` ``, `)`, `]`, `&`, `|` to the exclusion class covers the markup / URI cases. A trailing sentence period (`Password=secret.`) is better trimmed in the replacement than excluded from the class, since `.` can legitimately appear inside a value — excluding it would truncate the match and leave part of the value visible.

src/tools/PackageJsonGenerator/Helpers/CanonicalModelBuilder.cs
```diff
@@ -705,6 +705,7 @@ private static List<DocNode> CollectNodes(XElement element)
                     // Normalize whitespace within text runs (newlines → spaces, collapse runs)
                     text = text.Replace("\r\n", " ").Replace("\n", " ").Replace("\r", " ");
                     while (text.Contains("  ")) text = text.Replace("  ", " ");
+                    text = DocumentationSanitizer.RedactConnectionStringPasswords(text);
```
radical, line 708: [automated] These six call sites cover the doc-node and `<example>` paths, but a few text paths still reach the emitted JSON without passing through the sanitizer:

- `:84` — `Description = ExtractSummary(f)` for enum members flows through `ExtractPlainText` → `AppendPlainText`, which appends `<c>` / `<code>` content verbatim (`:1020`). 16 packages emit `enumMembers`.
- `:741` and `:843` — `<see href>` / `<a>` label text via `child.Value.Trim()`.

Nothing hits these today, so it isn't a live leak. The shape is the concern: sanitizing at scattered producer sites is exactly why the `<example>` paths were missed on the first pass, and the next text path added will miss it too. Worth sanitizing once at a chokepoint — e.g. inside `ExtractPlainText` and at `DocNode` / `DocExample` construction — rather than at each individual producer?

radical (commented):
[automated] Automated review notes on the connection-string password redaction change (3 inline comments below).

src/tools/PackageJsonGenerator/Helpers/DocumentationSanitizer.cs
radical: [automated] Done in radical/aspire.dev@458fab3c — the token is now `Placeholder`. The `{password}` expectations in `DocumentationSanitizerTests.cs` and the 4 `pkgs/*.json` files were updated to match; the pre-existing `{password}` doc-template tokens (e.g. the two in the Postgres file) were left untouched.

On `RedactConnectionStringPasswords_IsIdempotent`: I kept the existing assertion. With the bare `Placeholder` token the sanitizer does re-match its own output, but it re-emits an identical string, so "apply-once output is unchanged on a second pass" still holds and the test still passes. Happy to switch it to the explicit `apply-twice == apply-once` form you suggested if you'd prefer the stronger phrasing.

src/tools/PackageJsonGenerator/Helpers/DocumentationSanitizer.cs
radical: [automated] Fixed in radical/aspire.dev@8ca47ae7. Added `` ` ``, `)`, `]`, `&`, `|` to the exclusion class (the value is now capture group 2), so a trailing markdown/URI delimiter is no longer swallowed. A trailing `.` is trimmed off the value in the replacement callback rather than excluded from the class, exactly as you suggested — dotted values stay intact and only the sentence terminator is preserved.

Added unit tests covering the backtick / link-paren / bracket / ampersand / pipe and trailing-period cases; each fails on the old pattern and passes on the new one.

src/tools/PackageJsonGenerator/Helpers/CanonicalModelBuilder.cs
radical: [automated] Addressed in radical/aspire.dev@8ca47ae7 — `:84` (enum member `ExtractSummary`) and both `<see href>` / `<a>` label sites (`:741`, `:843`) now run through the sanitizer.

I wrapped those three producer sites rather than doing the chokepoint refactor, to keep the diff low-risk on this branch. I agree the chokepoint shape (sanitizing once inside `ExtractPlainText` and at `DocNode` / `DocExample` construction) is the better long-term design and is worth a follow-up — the surgical fix here closes the known paths without restructuring the builder.

Confirmed this changes no committed JSON (a grep of the current package set finds no real `password=` / `pwd=` values), so it's preventive. Added an end-to-end test: an enum member whose `<summary>` contains a connection string is redacted in the generated JSON — it fails if the `:84` wrap is reverted.

radical (approved):

PR #1420, opened 2026-08-03
Redesign the Aspire landing page experience

IEvangelist:
## Summary

- Redesign the English Aspire landing page with a GitHub-inspired narrative flow while preserving Aspire branding, routes, themes, and product claims.
- Refine the AppHost runtime story, live topology, AI context, environment states, dashboard carousel, integrations, testimonials, responsive motion, and reduced-motion behavior.
- Add accessible custom selects, localized shared header and footer actions, refreshed AspireShop dashboard media, and expanded unit and browser regression coverage.

Adds a link to the Aspire Reddit community and retains a link to the Microsoft-owned `dotnet/aspire` GitHub repository. No material affiliation with Reddit; the author is affiliated with Microsoft.

## Validation

- `pnpm lint`
- 283 unit tests passed across 23 files
- Full live Playwright matrix: 278 passed, 52 intentional skips across desktop, tablet, and mobile
- Post-rebase live smoke: 13 passed, 2 intentional viewport skips
- WCAG A/AA axe coverage, responsive reflow, dark theme, reduced motion, forced colors, keyboard operation, and horizontal overflow checks passed
- The three dynamic OG PNG checks are build-only and return 404 under `astro dev`; the endpoint is unchanged from `main` and the deployed PNG resolves successfully
- No production build was run, per request


alistairmatthews:
When I test this locally and try light mode, I find only some parts of the homepage turn light. Is this by design?
image

alistairmatthews:
Also, the light/dark mode controls are in the footer. I'd usually expect them in the header, especially for a long page like this.

alistairmatthews:
Really like the animation here, but the text is the same for each AI LLM with just the name substituted, which makes the animation seem like overkill. Is it possible to introduce a bit more variation in the text or think of things that are specific to each model?
image

alistairmatthews:
The alignment for the text "03 ONE APPLICATION MODEL" is different from other sections. Should be level with the top  of the title.
image

src/frontend/src/components/home/HomeEnvironment.astro
```diff
@@ -0,0 +1,1759 @@
...
+    ],
+  },
+  {
+    id: 'test',
+    label: 'Test',
+    eyebrow: 'Repeatable validation',
+    title: 'Test the topology you ship.',
+    description:
+      'Use the application model in automated tests and CI workflows so dependencies and relationships match the system developers run locally.',
+    action: { kind: 'link', label: 'Testing docs', href: '/testing/' },
```
alistairmatthews: This creates a link to a 404 error. I think it should be "/testing/overview/"

```suggestion
    action: { kind: 'link', label: 'Testing docs', href: '/testing/overview/' },
```

alistairmatthews:
This alignment is different from the previous sections, where the numbered subtitle is on the top left and the main title to its right. If this is intentional, no problem!
image

alistairmatthews:
This testimonial has incorrect punctuation. Should be a comma after "begin"?
image

alistairmatthews:
All nit-picks. Other than the above, this looks great.

IEvangelist:
> When I test this locally and try light mode, I find only some parts of the homepage turn light. Is this by design? image

Yes, this is by design.

IEvangelist:
> Also, the light/dark mode controls are in the footer. I'd usually expect them in the header, especially for a long page like this.

They've been in the footer for a while now...😬. Look at [aspire.dev](https://aspire.dev) today.

IEvangelist:
> This testimonial has incorrect punctuation. Should be a comma after "begin"? image

Yeah, I think the real issue here is that "skip" should be capitalized.

src/frontend/src/components/home/HomeEnvironment.astro
IEvangelist: Fixed in eac3d92b9: the Testing docs CTA now points to /testing/overview/. I also added a Playwright regression that deduplicates every same-origin homepage path, performs a real GET for each destination, and reports any non-successful status independently of the Starlight link validator.

IEvangelist:
> Really like the animation here, but the text is the same for each AI LLM with just the name substituted, which makes the animation seem like overkill. Is it possible to introduce a bit more variation in the text or think of things that are specific to each model? image

I thought about that, but I also don't want it to seem like one AI tool works differently with Aspire. The goal here was to convey that regardless of the AI tool, Aspire provides meaningful context not otherwise available.

maddymontaquila:
> > Really like the animation here, but the text is the same for each AI LLM with just the name substituted, which makes the animation seem like overkill. Is it possible to introduce a bit more variation in the text or think of things that are specific to each model? image
> 
> I thought about that, but I also don't want it to seem like one AI tool works differently with Aspire. The goal here was to convey that regardless of the AI tool, Aspire provides meaningful context not otherwise available.

Could just nuke the animation altogether?

maddymontaquila:
maybe "Deploy with the same app model" instead of "without losing"
image

maddymontaquila:
I feel the urge to click on these tooltips - no change needed but wanted to call it out 
image

maddymontaquila:
> > When I test this locally and try light mode, I find only some parts of the homepage turn light. Is this by design? image
> 
> Yes, this is by design.

It's a little too jarring IMO. Is there a softer background color to sub in? like a lavender?

maddymontaquila:
BAnner kinda blends in. what about one of the gradients from the brand book? 
image

maddymontaquila (approved):

PR #1430, opened 2026-08-05
Consume canonical TypeScript API exports

adamint:
Draft. Consumer half of microsoft/aspire#19032 — tracking #17608.

Today `AtsJsonGenerator` reconstructs TypeScript signatures on the site side: options-overload synthesis, C# inheritance borrowing, container-base inference, missing-declaration stubs. It's a second implementation of the TypeScript projection that lives in the wrong repo, and it drifts from what the generator actually emits. That's the bug behind #17608.

The plan is for the CLI to hand us the API surface for an exact `Name@Version` and for this repo to do nothing but validate, sort, dedupe, and write.

### What's in the draft so far

The schema v1 contract and the fixtures, so the switchover has something to land against:

- `src/schemas/typescript-api-export.ts` — parser/validator, typed loader, `concatenateDeclarations`
- `tests/fixtures/typescript-api-export/*.json` — unedited `aspire sdk export` output for `Aspire.Hosting` and `Aspire.Hosting.Redis`
- 16 unit tests, including one that concatenates every declaration fragment and shells out to `tsc --noEmit --strict` with `skipLibCheck` **off**

The fixtures are deliberately raw CLI output, not hand-tidied — if the producer drifts, these tests are what catches it. That already paid off once: the augmentation ownership test here caught `addRedis` shipping as `interface:DistributedApplicationBuilder` owned by `Aspire.Hosting.Redis`, which collides with core's item ID. Fixed on the producer side.

### Still to come on this branch

- Rewire `update-ts-api.ts` onto `aspire sdk export` and delete `AtsJsonGenerator`
- Reduce `generate-twoslash-types.ts` to validation + dedupe + write
- Drop the TypeScript half of `normalize-generated-api-data.ts`

Keeping it a draft until microsoft/aspire#19032 is in and there's a CLI on a feed CI can restore — the refresh workflow can't run against a `sdk export` that isn't published yet.

### Testing

328/328 unit tests, eslint clean.


src/frontend/src/schemas/typescript-api-export.ts
```diff
@@ -0,0 +1,320 @@
...
+    id: requireNonEmptyString(source, record.id, `${path}.id`),
+    kind: requireNonEmptyString(source, record.kind, `${path}.kind`),
+    name: requireNonEmptyString(source, record.name, `${path}.name`),
+    // A blank declaration means the producer failed to resolve a signature, which would otherwise
+    // surface as an empty code block on a published page.
+    declaration: requireNonEmptyString(source, record.declaration, `${path}.declaration`),
+    summary: optionalString(source, record.summary, `${path}.summary`),
+    remarks: optionalString(source, record.remarks, `${path}.remarks`),
+    deprecated: optionalString(source, record.deprecated, `${path}.deprecated`),
+    returnType: optionalString(source, record.returnType, `${path}.returnType`),
```
adamint, line 218: Could we model and validate `capabilityId` and `parameters` here too? Both canonical fixtures emit them, but `parseMember` drops them from the returned object. The current reference pages use `capabilityId` for identity/search and `parameters` for overload slugs and parameter docs, so the switchover would lose that data.

src/frontend/src/schemas/typescript-api-export.ts
```diff
@@ -0,0 +1,320 @@
...
+
+  const modules = requireArray(source, record.modules, 'modules').map((module, index) =>
+    parseModule(source, module, `modules[${index}]`),
+  );
+
+  const declarations = requireArray(source, record.declarations, 'declarations').map(
+    (declaration, index) => parseDeclaration(source, declaration, `declarations[${index}]`),
+  );
+
+  const seenItemIds = new Set<string>();
```
adamint, line 311: I think this only enforces item ID uniqueness inside one package document. The test below says the IDs must be unique across the full manifest because pages are keyed by them, but two individually valid exports can still collide and `concatenateDeclarations` never looks at modules. Could we enforce this in a manifest-level validator rather than only checking the two fixture values?

src/frontend/src/schemas/typescript-api-export.ts
```diff
@@ -0,0 +1,320 @@
...
+export function concatenateDeclarations(
+  documents: readonly TypeScriptApiExport[],
+): ConcatenatedDeclarations {
+  const byId = new Map<string, TypeScriptApiDeclaration>();
+
+  for (const document of documents) {
+    for (const declaration of document.declarations) {
+      const existing = byId.get(declaration.id);
+
+      if (existing !== undefined && existing.content !== declaration.content) {
```
adamint: Should this also reject an `owningAssembly` mismatch? Right now two declarations with the same ID/content but different owners are accepted, and the later document silently wins in `declarations`. That makes the returned metadata depend on document order and hides a producer ownership regression.

adamint (commented):
I found three contract gaps inline. I did not repeat the existing `tsc` stderr comment. CI is also blocked by the forbidden-words check because the raw `Aspire.Hosting.api.json` fixture contains `app host`; since the fixture is intentionally unedited, I think this either needs the producer output refreshed or the bypass label.

src/frontend/src/schemas/typescript-api-export.ts
adamint: Added both fields to the canonical member model. Parameters are validated field-by-field and preserved in producer order.

src/frontend/src/schemas/typescript-api-export.ts
adamint: Added manifest-level validation before declaration concatenation, so item ID collisions across individually valid package exports are rejected.

src/frontend/src/schemas/typescript-api-export.ts
adamint: This now rejects repeated declaration IDs when either content or ownership differs. Exact repeats still deduplicate.

src/frontend/tests/unit/typescript-api-export.vitest.test.ts
```diff
@@ -0,0 +1,262 @@
...
+    );
+
+    const tsc = fileURLToPath(new URL('../../node_modules/typescript/bin/tsc', import.meta.url));
+
+    let output = '';
+    try {
+      execFileSync(process.execPath, [tsc, '--project', tsconfig], { encoding: 'utf8' });
+    } catch (error) {
+      output = (error as { stdout?: string }).stdout ?? String(error);
+    }
```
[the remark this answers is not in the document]
adamint: The test now captures both stdout and stderr, then falls back to the error text when neither stream contains diagnostics.

src/frontend/src/schemas/typescript-api-export.ts
adamint: Added both fields to the canonical member model. Parameters are validated field-by-field and preserved in producer order.

src/frontend/tests/unit/typescript-api-export.vitest.test.ts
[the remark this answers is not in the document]
adamint: Good call, the test now captures both stdout and stderr and falls back to the error text when neither stream has diagnostics.

src/frontend/src/schemas/typescript-api-export.ts
adamint: Yep, this now rejects declaration IDs whose repeated fragments disagree on ownership as well as content. Exact repeats still deduplicate.

src/frontend/src/schemas/typescript-api-export.ts
adamint: Added a manifest-level validator and call it before concatenating declarations, so item ID collisions across individually valid package exports are rejected.

src/frontend/src/schemas/typescript-api-export.ts
```diff
@@ -0,0 +1,403 @@
...
+  const itemsById = new Map<string, string>();
+  const declarationsById = new Map<
+    string,
+    { declaration: TypeScriptApiDeclaration; packageName: string }
+  >();
+
+  for (const document of documents) {
+    for (const module of document.modules) {
+      for (const item of module.items) {
+        const firstPackage = itemsById.get(item.id);
```
IEvangelist: Could we key this by package identity (or `typeId`) instead of bare `item.id`? The producer does not make most item IDs globally unique: for example, options use [`Id = $"options:{interfaceName}"`](https://github.com/microsoft/aspire/blob/00d3ddb8ddadc3824d2086269ff05457cb497645/src/Aspire.Hosting.CodeGeneration.TypeScript/TypeScriptApiProjector.cs#L1470-L1471), with collision handling scoped to one package export. Two container integrations can therefore legitimately emit `options:WithDataVolumeOptions`, and this validator will reject the full manifest. The core + Redis test does not cover that case because core owns no matching options item. Please scope the key by package (or use the qualified `typeId`) and cover two integration exports.

src/frontend/src/schemas/typescript-api-export.ts
```diff
@@ -0,0 +1,403 @@
...
+export function concatenateDeclarations(
+  documents: readonly TypeScriptApiExport[],
+): ConcatenatedDeclarations {
+  validateTypeScriptApiManifest(documents);
+
+  const byId = new Map<string, TypeScriptApiDeclaration>();
+
+  for (const document of documents) {
+    for (const declaration of document.declarations) {
+      byId.set(declaration.id, declaration);
```
IEvangelist: The producer contract explicitly says declaration IDs are scoped to a package export and that declarations from separate exports [cannot be flattened into one global declaration set](https://github.com/microsoft/aspire/blob/00d3ddb8ddadc3824d2086269ff05457cb497645/src/Aspire.Hosting.CodeGeneration.TypeScript/TypeScriptApiModel.cs#L66-L71). This map/join does exactly that. Distinct IDs from Redis and MongoDB can still declare the same package-local TypeScript name (for example, `WithDataVolumeOptions`), causing TypeScript either to merge unrelated option bags or reject incompatible members. Please emit/type-check declarations per package, or isolate each package in its own module/namespace, rather than producing one global file.

IEvangelist (changes requested):
**Accuracy source:** `microsoft/aspire` `main` @ `00d3ddb8ddadc3824d2086269ff05457cb497645`

**Summary:** 23 claims checked: 17 verified, 4 verified with nuance, 2 contradicted, 0 unverifiable. The two contradictions are blocking and are called out inline. The doc-tester pass found no affected documentation routes; the PR head served successfully at `http://localhost:4322/` with 0 critical issues and 0 warnings.

## Phase A — Claim verification

<details>
<summary>Verified and verified-with-nuance claims</summary>

| ID | Verdict | Claim and evidence |
|---|---|---|
| C1 | Verified | `aspire sdk export --language typescript` exists (`SdkExportCommand.cs`: command/option definitions). |
| C2 | Verified | `schemaVersion` is literal `1` (`TypeScriptApiProjector.ExportSchemaVersion`). |
| C3 | Verified | Exported document language is lowercase `typescript` (`TypeScriptApiProjector.cs`). |
| C4 | Verified | Package identity contains non-empty `name` and `version` (`TypeScriptApiExportWriter.cs`). |
| C5 | Verified | Modules contain `name`, optional `summary`, and `items`; the producer currently emits one module. |
| C6 | Verified | Items emit `id`, `kind`, `name`, `declaration`, and `owningAssembly`. |
| C7 | Verified with nuance | The consumer permits missing `typeId`, although the producer requires and always emits it. |
| C8 | Verified | Item `examples` and `extends` are optional string arrays. |
| C9 | Verified with nuance | The producer also emits member-level `examples`; `parseMember` currently drops that field. |
| C10 | Verified | Member deprecation text is serialized under `deprecated`. |
| C11 | Verified | Parameters serialize as `name`, `type`, `optional`, and optional `summary`. |
| C12 | Verified | Declarations contain non-empty `id`, `content`, and `owningAssembly`. |
| C13 | Verified with nuance | The producer emits a top-level `generator` object; the consumer currently drops it. |
| C14 | Verified | Method declarations use the documented TypeScript signature form. |
| C15 | Verified | Item declarations use headers such as `export interface RedisResource`. |
| C16 | Verified | Extension methods are augmentation items with package-qualified augmentation IDs. |
| C19 | Verified | Exact repeated declarations are harmless within the consumer's dedupe behavior. |
| C20 | Verified | Fixture declaration IDs are unique and sorted for the tested inputs. |
| C21 | Verified with nuance | The concatenated core + Redis fixture type-checks, but this does not establish correctness for the full multi-package manifest. |
| C22 | Verified | The type-only re-export is erased and does not introduce `node:fs` into page bundles. |
| C23 | Verified | Fixture shape and field order match raw producer output, including `generator`. |

The producer evidence is in `TypeScriptApiModel.cs`, `TypeScriptApiExportWriter.cs`, `TypeScriptApiProjector.cs`, `AtsTypeScriptApiReferenceExporter.cs`, and `SdkExportCommand.cs` at the SHA above.
</details>

## Phase B — Doc-tester results

# Documentation Test Report

**Focus Area:** PR #1430 — no documentation content or rendered routes changed  
**Date:** 2026-08-25  
**Tester:** doc-tester agent  
**PR ref served:** `0fe13f5d23fe84a20c60815e167ee4cdeefe36c0` at `http://localhost:4322/`

| Category | Passed | Failed | Warnings |
|---|---:|---:|---:|
| Content Accuracy | 0 | 0 | 0 |
| Code Examples | 0 | 0 | 0 |
| CLI Commands | 0 | 0 | 0 |
| Links | 0 | 0 | 0 |

### Critical Issues

None. No documentation pages or sections changed in this PR.

### Warnings

None.

### Passed Checks

- The PR head served locally and `/` loaded successfully with the expected site title.
- The changed-file set contains only the TypeScript export schema, utility re-export, fixtures, and unit tests; there are no affected documentation routes to exercise.

### Recommendations

None for rendered documentation in this PR.

### Knowledge Gaps

None.

src/frontend/src/schemas/typescript-api-export.ts
adamint: Fixed. Declaration concatenation now accepts one package export, and the compiler proof writes core, Redis, and two colliding integrations as four separate TypeScript modules under one strict `tsconfig`. Nothing is flattened, wrapped, or rewritten by the site.

src/frontend/src/schemas/typescript-api-export.ts
adamint: Fixed. Bare item and declaration IDs are no longer validated across package exports. The regression builds two integration exports with the same `options:WithDataVolumeOptions` item ID, declaration ID, and TypeScript name, while keeping package/type identity and content distinct.

adamint:
Good call, fixed both. The import checks now validate each package as its own module instead of merging bare IDs/declarations globally, and the fixtures/schema now match the current export contract (`generator`, `typeId`, member examples, and closed `kind` validation). Focused tests, the frontend build, and CI are green.

PR #1448, opened 2026-08-07
Add declarative auto-expiry to docs announcement banner

IEvangelist:
## Summary

Adds **declarative auto-expiry** to the docs announcement banner, layered on top of the existing dismiss infrastructure. Version-update banners never got removed: if a reader saw one but never clicked dismiss, it lingered on every visit and became noise.

Two optional frontmatter fields, nested under `banner:`:

- **`banner.expiresOn`** — an absolute `YYYY-MM-DD` sunset date. Hides the banner for **everyone** after it passes (and skips rendering at build time once past). An editorial hard stop.
- **`banner.autoDismissAfterDays`** — a **per-reader** window. Auto-hides the banner _N_ days after that reader first saw it, tracked in `localStorage`, even if they never clicked dismiss.

Either condition (or an explicit dismiss) hides the banner, and both compose.

## Changes

- **`content.config.ts`** — extends Starlight's `banner` object with the two optional fields (+ JSDoc). Relies on Starlight 0.41's `extend` intersection, whose object merge preserves the nested keys.
- **`src/utils/banner-expiry.ts`** _(new)_ — pure, dependency-free decision helper (`resolveBannerVisibility` + guards), mirroring the existing `browser-storage.ts` pattern.
- **`Banner.astro`** — build-time short-circuit for a past `expiresOn`; new `firstSeen` storage key; client decision logic that persists first-seen and reuses the existing show/collapse paths (so the sidebar's `--aspire-banner-height` tracking keeps working untouched). Content hash stays content-only, so tweaking expiry never resets a reader's dismiss/first-seen state.
- **`banner-expiry.vitest.test.ts`** _(new)_ — 13 edge cases (boundaries, first-view persistence, in/out window, dismiss-wins, combined conditions), auto-discovered by `test:unit`.
- **Content rollout** — adds `autoDismissAfterDays: 14` to the current "Aspire 13.4" banners across all locales so they self-expire 14 days after a reader first sees them.
- **`.github/astro.instructions.md`** — documents the new behavior.

## Testing

- Pure logic verified 13/13 via the vitest spec (validated locally through Node's built-in `node:test` runner against the real source, since the sandbox blocks the npm registry).
- Schema behavior confirmed against the installed Starlight 0.41.3 source: `banner` is a non-strict `z.object` (unknown keys stripped, not rejected) and `extend` combines via `.and()` → `ZodIntersection`, whose merge preserves the nested `expiresOn` / `autoDismissAfterDays`.

> [!NOTE]
> The `autoDismissAfterDays: 14` default and the exact set of rolled-out pages are easy to tune — happy to adjust in review.


src/frontend/src/content.config.ts
```diff
@@ -10,6 +10,36 @@ export const collections = {
...
+              /** Banner HTML content (as in Starlight's built-in schema). */
+              content: z.string(),
+              /**
+               * Absolute sunset date. Once it passes, the banner is hidden
+               * for **everyone**, regardless of whether they dismissed it.
+               * Use `YYYY-MM-DD` in frontmatter (e.g. `2026-09-01`). If the
+               * date is already past at build time the banner isn't rendered
+               * at all.
+               */
+              expiresOn: z.coerce.date().optional(),
```
adamint: `expiresOn` never reaches `Banner.astro`. I added `expiresOn: 2020-01-01` to a banner and the emitted inline script still had `expiresOnMs = null`; the banner rendered normally. It looks like the Starlight schema intersection strips the coerced `Date`. Could we preserve this metadata outside the built-in banner object (or otherwise carry it through) and add a rendered-banner regression test?

src/frontend/src/components/starlight/Banner.astro
```diff
@@ -37,7 +58,16 @@ const dismissKey = bannerHash
         </button>
       </div>
 
-      <script is:inline define:vars={{ bannerId, dismissKey }}>
+      <script
+        is:inline
+        define:vars={{ bannerId, dismissKey, firstSeenKey, expiresOnMs, autoDismissAfterDays }}
+      >
+        // Mirror of resolveBannerVisibility() in src/utils/banner-expiry.ts —
```
adamint: This helper is only imported by its unit test; the production inline script reimplements it manually, so the tests can pass while the shipped behavior diverges. The browser run found two failures those unit tests could not see. Could we exercise the rendered `Banner.astro` path directly, or add a contract test that actually ties this inline implementation to the helper?

src/frontend/src/components/starlight/Banner.astro
```diff
@@ -92,10 +122,47 @@ const dismissKey = bannerHash
             dismissed = false;
           }
 
+          let firstSeenMs = null;
+
+          try {
+            const stored = window.localStorage.getItem(firstSeenKey);
+            const parsed = stored === null ? NaN : Number.parseInt(stored, 10);
```
adamint: `parseInt` accepts values like `12abc` and negative timestamps. I tested both in the browser; each hid the banner permanently and the stored value never self-healed. Could we require a positive integer timestamp no later than `now`, and otherwise treat this as a first view and overwrite it?

adamint (approved):
I found three things that I think need fixing before merge: `expiresOn` is stripped before it reaches the component, partially valid/negative first-seen values can permanently hide the banner, and the unit tests exercise a helper that production does not use. The browser run passed first view, day 13/day 15, manual dismissal, layout collapse, and client navigation. Approving because these are all local fixes.

src/frontend/src/content.config.ts
IEvangelist: Fixed in b777ce5. The expiry metadata is now emitted as **top-level** frontmatter keys (`bannerExpiresOn` / `bannerAutoDismissAfterDays`) instead of nested under `banner`.

Root cause: extra keys nested inside Starlight's built-in `banner` object (a plain `z.object({ content })`) get stripped before they reach the component, so `expiresOn` arrived as `null`. Top-level coerced dates survive the schema intersection intact — the same way the built-in `publishDate` reaches `MarkdownContent.astro` as a real `Date`.

I also added a rendered-`Banner.astro` regression test (`tests/unit/banner-render.vitest.test.ts`) that renders the real component and asserts `data-expires-on` / `data-auto-dismiss-days` actually appear in the emitted HTML, plus a build-time case proving a past `bannerExpiresOn` drops the banner entirely.

src/frontend/src/components/starlight/Banner.astro
IEvangelist: Fixed in b777ce5. The client controller is now a **bundled module** `<script>` that imports `resolveBannerVisibility` / `parseFirstSeen` from `banner-expiry.ts` directly, rather than re-implementing the decision logic inline. Because a bundled script can't use `define:vars`, per-banner config is passed via `data-*` attributes on the element and read from `dataset`.

Production and the unit tests now share a single source of truth, so shipped behavior can no longer diverge from what the tests cover. I also added `tests/unit/banner-render.vitest.test.ts`, which renders the real `Banner.astro` and asserts on its emitted output.

src/frontend/src/components/starlight/Banner.astro
IEvangelist: Fixed in b777ce5. Added `parseFirstSeen(raw, nowMs)`, which replaces the raw `parseInt`. It requires a positive integer timestamp no later than `now`: `^\d+$` (rejects `12abc`, `1.5`, `1e3`, `0x10`, empty), then `Number.isSafeInteger` + `> 0` (rejects `0` / negatives) + `<= nowMs` (rejects future/clock-skew values). Anything else returns `null`.

`resolveBannerVisibility` treats a `null` first-seen as a fresh first view and returns `persistFirstSeenMs: now`, so a corrupt/garbage stored value **self-heals** — it gets overwritten and the banner shows normally, instead of a bogus ~1970 timestamp hiding it permanently. New unit tests cover each rejected form plus the self-heal path.

PR #1461, opened 2026-08-11
Harden API data generation and fix C# deep links

### LINKED ISSUE #1462: C# API reference member deep links do not scroll to methods
## Description

C# API reference member-name deep links do not scroll to the requested method card.

## Reproduction

Open:

https://aspire.dev/reference/api/csharp/aspire.hosting.sqlserver/sqlserverbuilderextensions/methods/#withhostport

## Expected behavior

The page scrolls to and highlights the `WithHostPort` method while exact overload links remain unique and addressable.

## Actual behavior

The page loads without scrolling to the `WithHostPort` method because the generated member anchor is signature-specific and no stable member-name anchor exists.

## Proposed fix

PR #1461 adds stable member-name aliases, retains exact overload anchors, and covers both forms with browser regression tests.

IEvangelist:
## Summary

- make C# and TypeScript API-data generation exact, fail-closed, and transactional
- preserve package provenance, attribute payloads, DTO optionality, inheritance, and options shapes across generated artifacts
- add semantic regression gates and regenerate the checked-in API data
- normalize singular and plural `app host` terminology extracted from documentation comments and export-ignore reasons to `AppHost` and `AppHosts`, with a matching forbidden-words guard
- add stable C# member-name anchors so name and exact-overload deep links scroll to and highlight the expected method while overload links remain unique

## Validation

- PackageJsonGenerator: 38 tests passed
- AtsJsonGenerator: 5 tests passed
- focused frontend API/Twoslash tests: 27 tests passed
- AppHost terminology and Twoslash generation: 62 tests passed
- generated API semantic validation passed for 179 catalog packages, 109 TypeScript modules, 41 DTOs, and 327 handles
- all 348 Twoslash documentation blocks compiled without unexpected failures
- 8,281 C# members across 1,573 types produced collision-free anchors; `WithHostPort` resolves to `#withhostport`

Fixes #1462

### DIFF
--- a/src/frontend/scripts/aspire-terminology.ts
+++ b/src/frontend/scripts/aspire-terminology.ts
@@ -54,7 +54,7 @@ const terminologyRules: readonly TerminologyRule[] = [
     article: 'an',
   },
   {
-    term: String.raw`app${horizontalWhitespace}hosts`,
+    term: String.raw`(?<=polyglot${horizontalWhitespace})app${horizontalWhitespace}hosts`,
     replacement: 'AppHosts',
   },
   {
--- a/src/frontend/scripts/aspire-terminology.ts
+++ b/src/frontend/scripts/aspire-terminology.ts
@@ -54,6 +54,10 @@ const terminologyRules: readonly TerminologyRule[] = [
     article: 'an',
   },
   {
+    term: String.raw`app${horizontalWhitespace}hosts`,
+    replacement: 'AppHosts',
+  },
+  {
     term: String.raw`app${horizontalWhitespace}host`,
     replacement: 'AppHost',
   },
--- a/src/frontend/scripts/validate-generated-api-data.ts
+++ b/src/frontend/scripts/validate-generated-api-data.ts
@@ -97,7 +97,7 @@ export interface ValidationResult {
 
 interface ParsedInterface {
   parents: string[];
-  properties: Map<string, boolean>;
+  properties: Map<string, { optional: boolean; type: string }>;
 }
 
 interface AttributePayloadShape {
@@ -138,6 +138,28 @@ function shortTypeName(typeId: string): string {
 
 function camelCase(name: string): string {
   return name.length === 0 ? name : name[0].toLowerCase() + name.slice(1);
+}
+
+function normalizeTypeScriptType(typeName: string): string {
+  return typeName
+    .trim()
+    .replace(/[A-Za-z_][A-Za-z0-9_]*(?:[./][A-Za-z_][A-Za-z0-9_]*)+/g, (value) => {
+      const withoutAssembly = value.includes('/')
+        ? value.slice(value.lastIndexOf('/') + 1)
+        : value;
+      const parts = withoutAssembly.split('.');
+      return parts[parts.length - 1];
+    })
+    .replace(/\s+/g, '');
+}
+
+function normalizeClrType(typeName: string): string {
+  const slashIndex = typeName.indexOf('/');
+  return (slashIndex >= 0 ? typeName.slice(slashIndex + 1) : typeName)
+    .replace(/\s+/g, '')
+    .replace(/`\d+\[\[/g, '<')
+    .replace(/\],\[/g, ',')
+    .replace(/\]\]/g, '>');
 }
 
 function addUnique<T>(
@@ -289,8 +311,13 @@ function parseInterfaces(declarations: string): Map<string, ParsedInterface> {
     }
     const properties = new Map(previous?.properties ?? []);
     const body = declarations.slice(bodyStart, bodyEnd);
-    for (const property of body.matchAll(/^\s{2}([A-Za-z_][A-Za-z0-9_]*)(\?)?:/gm)) {
-      properties.set(property[1], property[2] === '?');
+    for (const property of body.matchAll(
+      /^\s{2}([A-Za-z_][A-Za-z0-9_]*)(\?)?:\s*(.+);$/gm
+    )) {
+      properties.set(property[1], {
+        optional: property[2] === '?',
+        type: property[3].trim(),
+      });
     }
 
     result.set(match[1], {
@@ -450,16 +477,24 @@ export function validateGeneratedApiData(input: ValidationInput): ValidationResu
     }
     for (const field of dto.fields ?? []) {
       const propertyName = camelCase(field.name);
-      const actualOptionality = declaration.properties.get(propertyName);
+      const property = declaration.properties.get(propertyName);
       if (!field.isOptional) {
         errors.push(
           `TypeScript DTO ${dto.name}.${propertyName} is required, but the SDK emits every DTO field as optional.`
         );
-      } else if (actualOptionality === undefined) {
+      } else if (!property) {
         errors.push(`Twoslash DTO ${dto.name} is missing property ${propertyName}.`);
-      } else if (!actualOptionality) {
+      } else if (!property.optional) {
         errors.push(
           `Twoslash DTO ${dto.name}.${propertyName} optionality does not match ts-modules metadata.`
+        );
+      }
+      if (
+        property &&
+        normalizeTypeScriptType(property.type) !== normalizeTypeScriptType(field.type)
+      ) {
+        errors.push(
+          `Twoslash DTO ${dto.name}.${propertyName} type ${property.type} does not match ts-modules metadata ${field.type}.`
         );
       }
     }
@@ -471,6 +506,24 @@ export function validateGeneratedApiData(input: ValidationInput): ValidationResu
     if (!declaration) {
       errors.push(`Twoslash handle ${handle.name} is missing its declaration.`);
       continue;
+    }
+    const matchingPackage = packageByIdentity.get(identity(module.package))?.data;
+    const packageType = (matchingPackage?.types ?? []).find(
+      (type) => type.fullName === handle.fullName
+    );
+    const generatedDirectBase = handle.baseTypeHierarchy?.[0];
+    if (packageType?.baseType) {
+      if (!generatedDirectBase) {
+        errors.push(
+          `TypeScript handle ${handle.name} is missing base hierarchy metadata for C# base type ${packageType.baseType}.`
+        );
+      } else if (
+        normalizeClrType(generatedDirectBase) !== normalizeClrType(packageType.baseType)
+      ) {
+        errors.push(
+          `TypeScript handle ${handle.name} base type ${generatedDirectBase} does not match C# metadata ${packageType.baseType}.`
+        );
+      }
     }
     const expectedParents = new Set(
       (handle.implementedInterfaces ?? [])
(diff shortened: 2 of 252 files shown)


src/frontend/src/components/api-reference/MemberCard.astro
```diff
@@ -447,6 +453,12 @@ if (formattedSigRaw && (member.kind === 'property' || member.kind === 'indexer')
       background 0.18s ease,
       box-shadow 0.18s ease;
   }
+  .mc-exact-anchor {
+    position: absolute;
+    inset-block-start: 0;
+    inset-inline-start: 0;
+    scroll-margin-top: calc(var(--sl-nav-height, 0px) + var(--sl-mobile-toc-height, 0px) + 1rem);
+  }
```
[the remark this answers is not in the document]
IEvangelist: Fixed in 6e8d3e9ef. Exact overload anchors now apply the target highlight to their parent member card, and the deep-link e2e test covers both the stable name anchor and the exact overload anchor.

src/frontend/scripts/aspire-terminology.ts
```diff
@@ -53,6 +53,10 @@ const terminologyRules: readonly TerminologyRule[] = [
     replacement: 'Aspire',
     article: 'an',
   },
+  {
+    term: String.raw`app${horizontalWhitespace}hosts`,
```
adamint: Could we avoid replacing bare `app hosts` without syntactic context? This rewrites verb phrases too: `the app hosts a service` becomes `the AppHosts a service`. The new tests only cover the plural noun. I think this needs a noun-context rule, or only the known generated phrases, plus a verb-usage regression test.

src/tools/AtsJsonGenerator/Helpers/AtsTransformer.cs
```diff
@@ -123,6 +123,9 @@ private static TsHandleTypeModel TransformHandle(AtsDumpHandleType h)
                 .Select(i => StripAssemblyPrefix(i.TypeId))
                 .OrderBy(i => i)
                 .ToList(),
+            BaseTypeHierarchy = h.BaseTypeHierarchy
+                .Select(i => StripAssemblyPrefix(i.TypeId))
```
adamint, line 127: This newly feeds base types through `StripAssemblyPrefix`, whose multi-argument parser drops every generic argument after the first. Reflection encodes arity-2 types as ``Type`2[[Arg1, Assembly],[Arg2, Assembly]]``; the regenerated Foundry data now contains ``AzureProvisionableAspireResourceWithParent`2`` with only its provisionable-resource argument and no parent-resource argument. Can we parse each bracketed argument independently, compare the hierarchy against the package metadata, and add an arity-2 test?

src/frontend/scripts/validate-generated-api-data.ts
```diff
@@ -0,0 +1,577 @@
...
+    }
+  }
+
+  for (const dto of selectedDtos.values()) {
+    const declaration = parsedInterfaces.get(dto.name);
+    if (!declaration) {
+      errors.push(`Twoslash DTO ${dto.name} is missing its declaration.`);
+      continue;
+    }
+    for (const field of dto.fields ?? []) {
```
adamint, line 478: Can we compare field types here as well? The gate currently checks only presence and optionality, so malformed metadata passes as a valid DTO shape. `src/data/ts-modules/Aspire.Hosting.13.4.6.json` already has `InteractionInput.Options` as `String]][]`, while the declaration is `string[]`, and this semantic validation still passes it.

src/frontend/src/utils/api-member-anchors.ts
```diff
@@ -0,0 +1,139 @@
...
+  const baseAnchors = members.map(memberSlug);
+  const baseCounts = new Map<string, number>();
+  for (const anchor of baseAnchors) {
+    baseCounts.set(anchor, (baseCounts.get(anchor) ?? 0) + 1);
+  }
+
+  const usedExactAnchors = new Set<string>();
+  const exactAnchors = baseAnchors.map((baseAnchor, index) => {
+    const initial =
+      (baseCounts.get(baseAnchor) ?? 0) > 1
```
adamint, line 68: Can we route every producer of member links through this resolver too? `MemberOverviewList.astro`, `[package]/index.astro`, and `csharp-api-markdown.ts` still emit `memberSlug(member)`. For the 12 real collision groups in the checked-in data, including the two `AddProject(...)` overloads, this changes both exact IDs to `base-hash`, so those producers link to the now-nonexistent raw `base` fragment. We should consume one stable mapping everywhere and preserve the previous base fragment as an alias when a collision appears.

src/frontend/src/data/pkgs/Aspire.Hosting.Azure.Kubernetes.13.4.6-preview.1.26319.6.json
```diff
@@ -2862,7 +2874,22 @@
...
           "hasGet": true,
-          "docs": {},
+          "docs": {
+            "summary": [
+              {
+                "kind": "text",
+                "text": " Gets the parent resource of type "
+              },
+              {
+                "kind": "typeparamref",
```
adamint: Can we substitute inherited generic parameters before publishing this documentation? `AzureKubernetesLoadBalancerResource` is not generic and this property returns `AzureKubernetesEnvironmentResource`, but the regenerated docs now say `Gets the parent resource of type T`. A regression test around inherited docs from a closed generic base should catch this.

src/tools/PackageJsonGenerator/generate-package-json.ps1
```diff
@@ -913,18 +944,26 @@ if (-not $Sequential -and $manifestEntries.Count -gt 1) {
         }
     }
 
-    if ($LASTEXITCODE -eq 0) {
-        $successCount = $manifestEntries.Count
-
-        foreach ($entry in $manifestEntries) {
+    foreach ($entry in $manifestEntries) {
+        if ($succeededPackages.Contains($entry.packageName)) {
```
adamint: Can batch mode use the same fail-closed success criteria as sequential mode? A `SUCCEEDED` marker increments success even when `$entry.output` is absent, and a nonzero process exit is only a warning. Full reconciliation can then delete the old final file because it is not staged and still publish. Please require a zero exit plus an output file for every success before continuing.

src/frontend/scripts/validate-generated-api-data.ts
```diff
@@ -0,0 +1,577 @@
...
+      const data = JSON.parse(fs.readFileSync(path.join(directory, fileName), 'utf8')) as T;
+      let baseline: T | undefined;
+      if (baselineDirectory) {
+        const relativePath = path
+          .relative(repoRoot, path.join(baselineDirectory, fileName))
+          .replaceAll(path.sep, '/');
+        const result = spawnSync('git', ['-C', repoRoot, 'show', `HEAD:${relativePath}`], {
+          encoding: 'utf8',
+        });
+        if (result.status === 0) {
```
adamint: Can we fail on unexpected `git show` errors here? Missing-at-HEAD is expected for a newly added file, but missing Git, an invalid HEAD, or permission errors silently disable every baseline comparison while the success output still says regressions were checked. I think we should only skip a confirmed absent path; otherwise surface stderr and fail validation.

adamint (dismissed):
I found several places where the new fail-closed generation and stable-anchor behavior do not hold yet. Details inline.

src/frontend/scripts/aspire-terminology.ts
IEvangelist: Addressed in 23b04c3f0. Plural normalization is now limited to the generated noun phrase after `polyglot`, with a regression proving `the app hosts a service` remains unchanged.

src/tools/AtsJsonGenerator/Helpers/AtsTransformer.cs
IEvangelist: Addressed in 23b04c3f0. Reflection generics now preserve every bracketed argument recursively, including nested generics and array arguments. The Foundry hierarchy is regenerated with both arity-2 arguments and is checked against package metadata.

src/frontend/scripts/validate-generated-api-data.ts
IEvangelist: Addressed in 23b04c3f0. DTO validation now compares exact field types as well as presence and optionality. The malformed `String]][]` source metadata was fixed at the generic formatter and both metadata and declarations now emit `KeyValuePair<string,string>[]`.

src/frontend/src/utils/api-member-anchors.ts
IEvangelist: Addressed in 23b04c3f0. All HTML, search, overview, Markdown, and cref link producers now consume the shared resolved-anchor map. Collision groups retain the old raw fragment as an alias, and Markdown destinations emit explicit exact and alias IDs.

src/frontend/src/data/pkgs/Aspire.Hosting.Azure.Kubernetes.13.4.6-preview.1.26319.6.json
IEvangelist: Addressed in 23b04c3f0. Inherited structured docs now substitute closed generic parameters before publication. Regression coverage verifies the concrete cref, and all affected Azure Kubernetes, Docker, Kubernetes, and K3s artifacts were regenerated.

src/tools/PackageJsonGenerator/generate-package-json.ps1
IEvangelist: Addressed in 23b04c3f0. Batch success now requires a zero process exit, exactly one terminal status, and an existing output file. Any violation marks the package failed and prevents staged output from being published.

src/frontend/scripts/validate-generated-api-data.ts
IEvangelist: Addressed in 23b04c3f0. HEAD paths are first checked with `git ls-tree`; only confirmed absence is skipped. Any `git show` or Git execution error now includes stderr and fails validation, with regression tests and a larger output buffer for package JSON.

src/frontend/src/data/pkgs/Aspire.Hosting.Azure.Kubernetes.13.4.6-preview.1.26319.6.json
adamint: This is still incomplete for inherited generic type-parameter docs. `Derived<U> : Base<U>` rewrites the node value from `T` to `U`, but the dictionary key remains `T`, so the generated JSON publishes docs for a nonexistent generic parameter. Could we remap or remove the `docs.typeParameters` keys too and add a type-level inheritdoc regression?

src/tools/PackageJsonGenerator/generate-package-json.ps1
adamint: The exact-one check still accepts duplicate identical markers because the parser stores statuses in `HashSet`s. Two `SUCCEEDED [Pkg]` lines collapse to one, `reportedStatusCount` stays 1, and the package publishes. Could we count status occurrences per package and reject any count other than one, with a duplicate-marker regression test?

adamint (approved):
The remaining two issues are straightforward fixes and are called out in the reopened threads. Focused frontend and generator tests, semantic validation, and live rendered anchor checks all pass. Approving.

PR #1467, opened 2026-08-11
docs: correct custom resource command CLI arguments to named options

### LINKED ISSUE #1463: Docs: custom resource command CLI arguments are named options (--flags), not positional
### Page

[`fundamentals/custom-resource-commands`](https://aspire.dev/fundamentals/custom-resource-commands/) → "Command arguments" / "Pass arguments from the CLI"

Source: [`custom-resource-commands.mdx`](https://github.com/microsoft/aspire.dev/blob/release/13.5/src/frontend/src/content/docs/fundamentals/custom-resource-commands.mdx)

### Problem

The page repeatedly states custom resource command arguments are passed on the CLI as **ordered positional values**:

- "Commands can declare input arguments that ... the CLI accepts as ordered positional values."
- InteractionInput `InputType` table → **CLI behavior** column: "Positional string value".
- "Pass arguments from the CLI ... supply argument values as ordered positional tokens after the command name":

```bash
aspire resource myservice send-message "Hello world" 3 --apphost MyApp.AppHost.csproj
```

> "`"Hello world"` maps to the first argument (`text`) and `3` maps to the second (`repeat`)."

But the CLI exposes these inputs as **named options** (`--flags`), not positional — see [`ResourceCommand.cs` L383-408](https://github.com/microsoft/aspire/blob/cfbf1c432e94dfe4a3261593eeef8c93913079bb/src/Aspire.Cli/Commands/ResourceCommand.cs#L383-L408). Each declared input becomes `new Option<...>($"--{optionName}")` (exposed as both exact-name and kebab-case). Leftover positional tokens are explicitly **not** bound:

> "Metadata-backed command inputs are options only. Any leftover token is forwarded as an unknown argument name so hosting-side validation reports it instead of binding it positionally."

Missing required inputs produce `Required option '--<name>' was not provided.` So the documented example fails: `"Hello world"` and `3` are treated as unrecognized tokens rather than mapped to `text` / `repeat`.

### Correct usage

```bash
aspire resource myservice send-message --text "Hello world" --repeat 3 --apphost MyApp.AppHost.csproj
```

(kebab-case aliases are also accepted, e.g. `--log-level` for an input named `LogLevel`.)

### Suggested fix

Update the "Command arguments" intro, the InteractionInput **CLI behavior** column ("Positional string value" / "Positional string parsed as number"), and the "Pass arguments from the CLI" section to show named `--<name>` options (with kebab-case aliases) and the required-option error, instead of positional tokens.

### Provenance

aspire.dev `release/13.5` (`custom-resource-commands.mdx`); aspire `release/13.5` (`ResourceCommand.cs`). Confirmed against the 13.5 candidate CLI `13.5.0-pr.17553`.

IEvangelist:
Fixes #1463.

The `fundamentals/custom-resource-commands` page described custom resource command CLI arguments as **ordered positional values**, but the CLI actually exposes each declared input as a named `--<name>` option. The documented example (`send-message "Hello world" 3`) fails in practice because those tokens are treated as unknown, not bound positionally.

## Changes

- **"Command arguments" intro** and the **`InteractionInput` properties** lead-in now say the CLI accepts named `--<name>` options.
- **`InputType` table (CLI behavior column)** — replaced "Positional string value" / "Positional string parsed as number" with the actual option forms (`--<name> <value>`, and `--<name>` flag for booleans).
- **"Pass arguments from the CLI" section** — the example now uses `--text "Hello world" --repeat 3`, notes that argument order doesn't matter, documents the kebab-case aliases (e.g. an input named `LogLevel` accepts both `--LogLevel` and `--log-level`), and shows the `Required option '--<name>' was not provided.` error.

## Source of truth

Verified against [`ResourceCommand.cs`](https://github.com/microsoft/aspire/blob/release/13.5/src/Aspire.Cli/Commands/ResourceCommand.cs) on `release/13.5`:

- Each input becomes `new Option<...>($"--{optionName}")`, where `optionName` is the kebab-cased `Name`, with the exact `Name` added as an alias.
- Missing required inputs produce `Required option '--<name>' was not provided.`.
- Leftover positional tokens are forwarded as unknown argument names, not bound positionally.

All angle brackets in the new copy stay inside inline-code spans / code fences, so MDX doesn't parse them as JSX (matching the existing `IReadOnlyList<...>` table cell).

Targets `release/13.5` as requested in the issue.

### DIFF
--- a/src/frontend/src/content/docs/fundamentals/custom-resource-commands.mdx
+++ b/src/frontend/src/content/docs/fundamentals/custom-resource-commands.mdx
@@ -1139,7 +1139,7 @@ The `InputType` enum controls how the dashboard renders the field and how built-
 |-------|-----------------|--------------|---------------------|
 | `Text` | Single-line text box | `--<name> <value>` string option | `Required`, `MaxLength` |
 | `Number` | Number input | `--<name> <value>` parsed as number | Must be a valid number |
-| `Boolean` | Checkbox | `--<name>` flag (`true` when present) | Must be `true` or `false` |
+| `Boolean` | Checkbox | `--<name>` (sets `true`) or `--<name> <value>` | Must be `true` or `false` |
 | `Choice` | Drop-down list | `--<name> <value>` matching one of `Options` | Must be a member of `Options` |
 | `SecretText` | Masked text box | `--<name> <value>` string option | `Required`, `MaxLength` |
 
@@ -1190,8 +1190,16 @@ aspire resource myservice send-message --text "Hello world" --repeat 3 --apphost
 
 In the example above, `--text` sets the `text` argument and `--repeat` sets the `repeat` argument. Each option name matches the input's `Name`, and a kebab-case alias is also accepted for multi-word names. For example, an input named `LogLevel` accepts both `--LogLevel` and `--log-level`.
 
+If an argument's option name matches one of the Aspire CLI's own options (for example the global `--log-level` / `-l` option, or `--apphost`), the CLI binds it during its own parsing pass and it never reaches the command. Separate the command's options from the CLI's options with a `--` delimiter to force them through:
+
+```bash title="Terminal"
+aspire resource myservice configure -- --log-level Debug
+```
+
+The command's `--help` output flags any names that collide with a CLI option and shows the `--` form to use.
+
 <Aside type="note">
-The CLI passes all option values as strings; the AppHost performs type coercion and built-in validation before invoking the command callback. If a required argument is missing, the CLI reports `Required option '--<name>' was not provided.` and doesn't invoke the callback. Any value that fails validation returns a structured error the same way.
+The CLI parses and validates values before forwarding them to the AppHost as strings: it parses `Number` and `Boolean` inputs and enforces the built-in rules (such as `Required` and `Choice` membership). If a required argument is missing, the CLI reports `Required option '--<name>' was not provided.` and doesn't invoke the callback. Any value that fails validation returns a structured error the same way.
 </Aside>
 
 ## Argument validation
--- a/src/frontend/src/content/docs/fundamentals/custom-resource-commands.mdx
+++ b/src/frontend/src/content/docs/fundamentals/custom-resource-commands.mdx
@@ -1041,7 +1041,7 @@ fi
 
 ## Command arguments
 
-Commands can declare input arguments that the dashboard renders as a prompt dialog before execution and that the CLI accepts as ordered positional values. Arguments are defined by setting `CommandOptions.Arguments` to an array of `InteractionInput` objects.
+Commands can declare input arguments that the dashboard renders as a prompt dialog before execution and that the CLI accepts as named `--<name>` options. Arguments are defined by setting `CommandOptions.Arguments` to an array of `InteractionInput` objects.
 
 <Tabs syncKey="aspire-lang">
 <TabItem id="csharp" label="C#">
@@ -1121,7 +1121,7 @@ await builder.build().run();
 
 ### InteractionInput properties
 
-Each `InteractionInput` object in the `Arguments` array configures one input field in the dashboard prompt and one positional argument for the CLI:
+Each `InteractionInput` object in the `Arguments` array configures one input field in the dashboard prompt and one named `--<name>` option for the CLI:
 
 | Property | Type | Description |
 |----------|------|-------------|
@@ -1137,11 +1137,11 @@ The `InputType` enum controls how the dashboard renders the field and how built-
 
 | Value | Dashboard widget | CLI behavior | Built-in validation |
 |-------|-----------------|--------------|---------------------|
-| `Text` | Single-line text box | Positional string value | `Required`, `MaxLength` |
-| `Number` | Number input | Positional string parsed as number | Must be a valid number |
-| `Boolean` | Checkbox | `true` or `false` string | Must be `true` or `false` |
-| `Choice` | Drop-down list | String must match one of `Options` | Must be a member of `Options` |
-| `SecretText` | Masked text box | Positional string value | `Required`, `MaxLength` |
+| `Text` | Single-line text box | `--<name> <value>` string option | `Required`, `MaxLength` |
+| `Number` | Number input | `--<name> <value>` parsed as number | Must be a valid number |
+| `Boolean` | Checkbox | `--<name>` flag (`true` when present) | Must be `true` or `false` |
+| `Choice` | Drop-down list | `--<name> <value>` matching one of `Options` | Must be a member of `Options` |
+| `SecretText` | Masked text box | `--<name> <value>` string option | `Required`, `MaxLength` |
 
 ### Read argument values at runtime
 
@@ -1182,16 +1182,16 @@ Use `requiredValue` for arguments declared with `required: true` to get a non-nu
 
 ### Pass arguments from the CLI
 
-When invoking a command from the CLI, supply argument values as ordered positional tokens after the command name. The CLI maps them onto the declared `Arguments` array by position:
+When invoking a command from the CLI, supply argument values as named `--<name>` options after the command name. Each declared argument is exposed as its own option, so the order you pass them in doesn't matter:
 
 ```bash title="Terminal"
-aspire resource myservice send-message "Hello world" 3 --apphost MyApp.AppHost.csproj
+aspire resource myservice send-message --text "Hello world" --repeat 3 --apphost MyApp.AppHost.csproj
 ```
 
-In the example above, `"Hello world"` maps to the first argument (`text`) and `3` maps to the second (`repeat`).
+In the example above, `--text` sets the `text` argument and `--repeat` sets the `repeat` argument. Each option name matches the input's `Name`, and a kebab-case alias is also accepted for multi-word names. For example, an input named `LogLevel` accepts both `--LogLevel` and `--log-level`.
 
 <Aside type="note">
-The CLI passes all positional values as strings; the AppHost performs type coercion and built-in validation before invoking the command callback. If a required argument is missing or a value fails validation, the CLI receives a structured error instead of invoking the callback.
+The CLI passes all option values as strings; the AppHost performs type coercion and built-in validation before invoking the command callback. If a required argument is missing, the CLI reports `Required option '--<name>' was not provided.` and doesn't invoke the callback. Any value that fails validation returns a structured error the same way.
 </Aside>
 
 ## Argument validation


src/frontend/src/content/docs/fundamentals/custom-resource-commands.mdx
```diff
@@ -1137,11 +1137,11 @@ The `InputType` enum controls how the dashboard renders the field and how built-
...
 | Value | Dashboard widget | CLI behavior | Built-in validation |
 |-------|-----------------|--------------|---------------------|
-| `Text` | Single-line text box | Positional string value | `Required`, `MaxLength` |
-| `Number` | Number input | Positional string parsed as number | Must be a valid number |
-| `Boolean` | Checkbox | `true` or `false` string | Must be `true` or `false` |
-| `Choice` | Drop-down list | String must match one of `Options` | Must be a member of `Options` |
-| `SecretText` | Masked text box | Positional string value | `Required`, `MaxLength` |
+| `Text` | Single-line text box | `--<name> <value>` string option | `Required`, `MaxLength` |
+| `Number` | Number input | `--<name> <value>` parsed as number | Must be a valid number |
+| `Boolean` | Checkbox | `--<name>` flag (`true` when present) | Must be `true` or `false` |
```
adamint: Could we document this as an optional Boolean value rather than a presence-only flag? The CLI configures these as `Option<bool>` with `ArgumentArity.ZeroOrOne`, so both `--verbose` and `--verbose false` are valid. Saying presence always means `true` is incorrect.

src/frontend/src/content/docs/fundamentals/custom-resource-commands.mdx
```diff
@@ -1182,16 +1182,16 @@ Use `requiredValue` for arguments declared with `required: true` to get a non-nu
...
-When invoking a command from the CLI, supply argument values as ordered positional tokens after the command name. The CLI maps them onto the declared `Arguments` array by position:
+When invoking a command from the CLI, supply argument values as named `--<name>` options after the command name. Each declared argument is exposed as its own option, so the order you pass them in doesn't matter:
 
 ```bash title="Terminal"
-aspire resource myservice send-message "Hello world" 3 --apphost MyApp.AppHost.csproj
+aspire resource myservice send-message --text "Hello world" --repeat 3 --apphost MyApp.AppHost.csproj
 ```
 
-In the example above, `"Hello world"` maps to the first argument (`text`) and `3` maps to the second (`repeat`).
+In the example above, `--text` sets the `text` argument and `--repeat` sets the `repeat` argument. Each option name matches the input's `Name`, and a kebab-case alias is also accepted for multi-word names. For example, an input named `LogLevel` accepts both `--LogLevel` and `--log-level`.
```
adamint, line 1191: Do we need to call out the `--` delimiter for names that collide with Aspire options? `ResourceCommand` first parses `--apphost`, `--include-hidden`, etc., then builds dynamic options from only `UnmatchedTokens`. An input named `AppHost` will not reach the command unless it is after `--`, but the current wording says every name is accepted directly.

src/frontend/src/content/docs/fundamentals/custom-resource-commands.mdx
```diff
@@ -1182,16 +1182,16 @@ Use `requiredValue` for arguments declared with `required: true` to get a non-nu
...
 
 <Aside type="note">
-The CLI passes all positional values as strings; the AppHost performs type coercion and built-in validation before invoking the command callback. If a required argument is missing or a value fails validation, the CLI receives a structured error instead of invoking the callback.
+The CLI passes all option values as strings; the AppHost performs type coercion and built-in validation before invoking the command callback. If a required argument is missing, the CLI reports `Required option '--<name>' was not provided.` and doesn't invoke the callback. Any value that fails validation returns a structured error the same way.
```
adamint: I think this assigns too much to the AppHost. The CLI validates required, disabled, and choice inputs and parses Boolean and number values before serializing the parsed values back to strings for the AppHost. Could we say the AppHost receives strings, while CLI validation and coercion happen first?

adamint (dismissed):
A few behavior details in the updated CLI guidance are still inaccurate. Details inline.

IEvangelist (commented):
## Docs-accuracy review — verdict: **COMMENT** ✅ (accurate; two wording nuances)

**Phase A source of truth:** `microsoft/aspire` @ `release/13.5` (SHA `1d922813c0f9bc14f7d817c817a8ab46b288bb44`), file `src/Aspire.Cli/Commands/ResourceCommand.cs`.
**Claims extracted:** 14 non-narrative → **12 verified**, **2 verified‑with‑nuance**, **0 contradicted**, **0 unverifiable**.
**Phase B (doc-tester, blind):** exercised `/fundamentals/custom-resource-commands/` (intro, *InteractionInput properties* / `InputType` table, *Pass arguments from the CLI*). **0 critical, 1 minor** styling note. Page renders correctly and no stale "positional" wording remains.

The core correction is right: custom resource command arguments are exposed by the CLI as **named `--<name>` options**, not ordered positional tokens. `ResourceCommand.CreateCommandArgumentOption` builds a `System.CommandLine` `Option` per argument (`new Option<...>($"--{optionName}")`) and the parser explicitly forwards leftover tokens as *unknown* names rather than binding them positionally (`ResourceCommand.cs:349-355`: *"Metadata-backed command inputs are options only… instead of binding it positionally."*). Boolean → flag, Number → `Option<double?>`, Choice/Text/SecretText → validated string option, and the `Required option '--<name>' was not provided.` message all match the source exactly.

---

### Phase A — Claim verification

#### Nuances worth considering (both `verified-with-nuance`, non-blocking)

**1. Which spelling is the option's canonical name vs. the alias is inverted.**
> "Each option name matches the input's `Name`, and a kebab-case alias is also accepted for multi-word names. For example, an input named `LogLevel` accepts both `--LogLevel` and `--log-level`."

The user-facing takeaway is correct — both `--LogLevel` and `--log-level` are accepted. But the direction is reversed relative to the source: the **kebab-case** form is the *canonical* option name and the exact `Name` is added as the *alias*, not the other way around:

- `ResourceCommand.cs:387` — `var optionName = ToKebabCase(argument.Name);` → the option is created as `--{optionName}` (kebab).
- `ResourceCommand.cs:444-448` — `var exactName = $"--{argument.Name}"; if (exactName != "--{optionName}") option.Aliases.Add(exactName);` → the exact `Name` is the *added alias*.
- `ResourceCommand.cs:855-858` and `:888` — `--help`/usage renders the **kebab-case** name as the canonical option label.

So for `LogLevel`, `--help` shows `--log-level` (canonical) and `--LogLevel` works as an alias. Consider rewording to something like: *"Each argument is exposed as an option whose canonical name is the kebab-case form of the input's `Name` (e.g. `--log-level`); the exact `Name` (`--LogLevel`) is also accepted as an alias."* For single-word names (`text`, `repeat`, `url`) exact == kebab, so the current example command is unaffected.

**2. "the AppHost performs … built-in validation" is slightly under-stated for the CLI.**
> "The CLI passes all option values as strings; the AppHost performs type coercion and built-in validation before invoking the command callback."

Values are indeed forwarded as strings (`ResourceCommand.cs:334-346` stringifies bool/number/string into the arguments payload). But the **CLI itself** also performs some built-in validation before anything reaches the AppHost — required-option presence (`:294-308`), choice membership (`:421-431`), and numeric parsing via `Option<double?>` (`:398`). The very next sentence in this PR (required-option error emitted by the CLI) is consistent with that, so this is only a small simplification. Optional: *"…the AppHost performs the remaining type coercion and validation…"* or note that basic required/choice/number checks happen CLI-side.

<details>
<summary>Verified claims (12) with evidence — <code>microsoft/aspire</code> @ <code>release/13.5</code>, <code>src/Aspire.Cli/Commands/ResourceCommand.cs</code></summary>

| # | Claim | Verdict | Evidence |
|---|-------|---------|----------|
| 1 | "Command arguments" intro: CLI accepts args as named `--<name>` options | verified | `:287-292`, `:394/398/404`, `:349-355` |
| 2 | *InteractionInput properties* intro: each input → "one named `--<name>` option for the CLI" | verified | `:287-292`, `:404` |
| 3 | `Text` → `--<name> <value>` string option | verified | `:404` (`Option<string?>`) |
| 4 | `Number` → `--<name> <value>` parsed as number | verified | `:398` (`Option<double?>`) |
| 5 | `Boolean` → `--<name>` flag (`true` when present) | verified | `:394-397` (`Option<bool>`), `:412-416` (`ArgumentArity.ZeroOrOne`), `:334-336`, help shows no value placeholder `:856-858` |
| 6 | `Choice` → `--<name> <value>` matching one of `Options` | verified | `:404` + choice validator `:421-431` |
| 7 | `SecretText` → `--<name> <value>` string option | verified | `:404` (non-bool/number → string) |
| 8 | CLI section intro: named options, "each declared argument is exposed as its own option, so the order … doesn't matter" | verified | `:287-292` (one option per arg; named parsing is order-independent) |
| 9 | Example `aspire resource myservice send-message --text "Hello world" --repeat 3 --apphost MyApp.AppHost.csproj` | verified | `--text`/`--repeat` = `--{ToKebabCase(name)}` (`:387`,`:404`); `--apphost` is a real option (`:43`) |
| 10 | "`--text` sets the `text` argument and `--repeat` sets the `repeat` argument" | verified | `:404`, `:343-345` |
| 11 | Required-missing → CLI reports `` Required option '--<name>' was not provided. `` | verified | `:301-303` (exact string), name built at `:298` (a plural variant exists at `:305-307`, not documented — fine) |
| 12 | "Any value that fails validation returns a structured error the same way" | verified | `:315-324` returns joined error message; validators `:421-441` |

</details>

---

### Phase B — Doc-tester results (blind, `playwright` against local `pnpm dev` build of this PR)

**Route exercised:** `http://localhost:<dev>/fundamentals/custom-resource-commands/` — *Command arguments*, *InteractionInput properties* + `InputType` table, and *Pass arguments from the CLI*.

**🔴 Critical issues:** none.

**🟡 Warnings (1, minor / not introduced by this PR):**
- The inline `` `--<name>` `` (and `` `--text` ``, `` `--log-level` `` …) code chips visually render the leading `--` as a single long dash because the site's monospace font (**Fira Code**) ligates double hyphens (`font-variant-ligatures: normal`, so `calt` stays on). The underlying DOM text and copy value are correct — two U+002D hyphens (confirmed: `hasEmDash: false`, `doubleHyphen: true`). This is **site-wide styling**, not a defect in this change, but this PR adds many inline `--` chips so the ligature is more visible here. A reader who only skims the prose *might* misread `--` as an em-dash, though the fenced `Terminal` example disambiguates it. Purely cosmetic; flagging for awareness, not as a change request.

**✅ Passed checks:**
- Page returns HTTP 200 and renders cleanly (tabs, tables, code blocks, `<Aside>` note all intact).
- "Command arguments" and *InteractionInput properties* intros both render with the new named‑option wording.
- `InputType` table renders all five rows/columns correctly — the literal `<name>` / `<value>` angle brackets inside backticks did **not** break the markdown table or get parsed as JSX/MDX.
- *Pass arguments from the CLI*: prose, `Terminal` code block, and the `<Aside type="note">` all render; the example command and the follow-up explanation are internally consistent (`--text`/`--repeat` in the example match the prose).
- **Consistency:** the page contains **zero** remaining "positional" / "ordered positional" / "by position" / "maps to the first argument" text — no stale contradictions left behind by the edit.

**Knowledge gaps:** none blocking. The `LogLevel` example is a helpful touch for explaining multi-word naming; a new user can follow the section end-to-end.

**Console note:** 7 console errors were observed, all Vite dev-server artifacts (`504 Outdated Optimize Dep` for the Astro dev toolbar modules: `astro`, `xray`, `audit`, `settings`, `toolbar`, `ui-library`). These are dev-only and unrelated to the documentation content.

---

*Automated docs-accuracy review (doc-pr-reviewer). Phase A verified prose against `microsoft/aspire` source; Phase B validated the rendered page blind to source. No contradictions and no critical rendering issues found — only the two optional wording nuances above.*

src/frontend/src/content/docs/fundamentals/custom-resource-commands.mdx
IEvangelist: You're right. Fixed in 72d648c8 — the Boolean row now documents `--<name>` (sets `true`) or `--<name> <value>` instead of a presence-only flag, and the validation column still notes the explicit value must be `true` or `false`.

src/frontend/src/content/docs/fundamentals/custom-resource-commands.mdx
IEvangelist: Added in 72d648c8. There's now a note that when an argument's option name matches one of the CLI's own options (for example the global `--log-level` / `-l` or `--apphost`), it must be passed after a `--` delimiter, with an example — and a mention that the command's `--help` surfaces the hint.

src/frontend/src/content/docs/fundamentals/custom-resource-commands.mdx
IEvangelist: Good call. Fixed in 72d648c8 — the note now says the CLI parses `Number`/`Boolean` and enforces the built-in validation (such as `Required` and `Choice`), then forwards string values to the AppHost.

PR #1482, opened 2026-08-12
Pin GitHub Actions to full-length commit SHAs

danfiedler-msft:
## Summary

- Actions remain pinned to full-length SHAs with standardized inline `# vX.Y.Z` comments.
- Dependabot `cooldown: { default-days: 7 }` now applies to all ecosystems in `dependabot.yml`: npm, nuget, and github-actions. Previously, only github-actions had this cooldown.
- `update-release-branch.yml` was already SHA-pinned; only its version comment was standardized from `# v3` to `# v3.0.0`.


.github/dependabot.yml
IEvangelist, line 1: Could you please also add the default cooldown for the other package-ecosystems as well?

.github/workflows/update-release-branch.yml
IEvangelist, line 1: This only actually changes the version comment, not the pinned SHA. The PR description is misleading, as we were already pinning to full-length commit SHAs.

.github/dependabot.yml
IEvangelist: Done — added `cooldown: default-days: 7` to the `npm` and `nuget` ecosystems as well, so all three ecosystems now share the 7-day cooldown.

.github/workflows/update-release-branch.yml
IEvangelist: You're right — this file was already SHA-pinned; only the version comment was standardized (`# v3` → `# v3.0.0`). I've corrected the PR description so it no longer implies new SHA pinning was introduced here.

IEvangelist (approved):

PR #1490, opened 2026-08-13
Harden CustomSelect against a stale open-time scroll dismissing the menu

IEvangelist:
## Summary

Hardens the footer language `CustomSelect` against a stale, open-time `scroll` event that could dismiss the menu the same frame it opened. This was surfacing as a flaky E2E test on the **mobile-chromium** project and intermittently red-lighting otherwise-unrelated chore PRs (e.g. #1487, [run 31721890999](https://github.com/microsoft/aspire.dev/actions/runs/31721890999)).

## Root cause

`CustomSelect.astro` registers a capture-phase `window` `scroll` listener that dismisses the menu on any page scroll, so the `position: fixed` listbox never visually detaches from its trigger. That is the desired behavior for a real scroll — but on small touch viewports, Playwright's `.click()` first has to scroll the far-below-the-fold footer control into view. On the Pixel 7 viewport that scroll event can be delivered a frame **after** `open()` runs, even though the page has already settled, so the freshly opened menu was immediately closed. The `language selector stays open while its listbox is scrolled` test then failed waiting for the listbox to be visible. Desktop (1440×900) and tablet (iPad Pro 11) scroll far less on the way to the footer and never hit the race.

## Fix

- Capture `window.scrollX/scrollY` when the menu opens.
- In the scroll handler, ignore events that report that same already-settled offset (with 2px of slop for sub-pixel rounding on high-DPI mobile) — only dismiss once the page actually moves underneath the menu.
- Split the combined viewport handler so `resize` still closes unconditionally while `scroll` takes the guarded path. The existing "scroll originates inside the listbox" guard is preserved.

The change is deterministic rather than timing-based (no added waits/retries), so it hardens the behavior for real users too, not just the test.

## Testing

- `astro sync` clean (no type errors in the client script).
- `pnpm test:unit:components` — 62/62 pass.
- Added a deterministic regression assertion to `ui-regressions.spec.ts` that dispatches a same-offset `window` `scroll` after opening and asserts the menu stays open. This fails on the old handler and passes with the fix, on every project. Full E2E runs in CI.

### DIFF
--- a/src/frontend/src/components/CustomSelect.astro
+++ b/src/frontend/src/components/CustomSelect.astro
@@ -148,6 +148,11 @@ const hasIcon = Astro.slots.has('icon');
       );
       let typeahead = '';
       let typeaheadTimer = 0;
+      // Scroll offset captured when the menu opens, used to distinguish a real
+      // page scroll (dismiss) from a stale scroll event still in flight from the
+      // gesture that opened the menu (keep open). See handleWindowScroll below.
+      let openScrollX = 0;
+      let openScrollY = 0;
 
       const isOpen = () => root.hasAttribute('data-open');
 
@@ -233,6 +238,8 @@ const hasIcon = Astro.slots.has('icon');
       function open() {
         if (isOpen()) return;
 
+        openScrollX = window.scrollX;
+        openScrollY = window.scrollY;
         document.dispatchEvent(new CustomEvent(openEventName, { detail: root }));
         listbox.hidden = false;
         if (typeof listbox.showPopover === 'function') {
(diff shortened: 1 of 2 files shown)


src/frontend/src/components/CustomSelect.astro
```diff
@@ -350,14 +357,29 @@ const hasIcon = Astro.slots.has('icon');
...
+        //      opened the control. Tapping the footer selector scrolls it into
+        //      view first; on small touch viewports that scroll event can be
+        //      delivered a frame later — after open() — even though the page has
+        //      already settled. Such stale events report the same offset we
+        //      captured when opening, so only dismiss once the page actually
+        //      moves underneath the menu (with a couple of pixels of slop to
+        //      absorb sub-pixel rounding on high-DPI mobile viewports).
         const target = event.target;
         if (target instanceof Node && listbox.contains(target)) return;
+        if (
```
adamint, line 394: Could we scope this guard to document/window scrolls? Right now any element scroller leaves `window.scrollX/Y` unchanged, so scrolling `.modal-content` around a `CustomSelect` never closes the fixed-position listbox and it detaches from the trigger. I think element scrolls should keep closing the menu; only the stale window/document event should be ignored.

adamint (approved):
The stale window-scroll path works in real Chromium across desktop, mobile, and tablet projects, with 21 focused scenarios passing. I left one must-fix inline: ancestor element scrolls now keep the fixed listbox open and detached because `window.scrollX/Y` do not move. That looks fixable in place, so approving with the comment.

src/frontend/src/components/CustomSelect.astro
IEvangelist: Good catch. Scoped the stale-offset guard to page/document/window scrolls only. `handleWindowScroll` now checks the event `target` against `document`, `window`, `document.scrollingElement`, `document.documentElement`, and `document.body`; only those are eligible for the same-offset short-circuit. Any other (ancestor element) scroll target falls through to `close()`, so scrolling something like `.modal-content` around a `CustomSelect` dismisses the menu instead of leaving the fixed listbox detached from its trigger. The internal-listbox guard and the stale same-offset window case are unchanged.

Also added a deterministic E2E regression that opens the footer language selector and dispatches a `scroll` on an ancestor element, asserting the menu closes. Pushed in d1c3b2e8.

src/frontend/src/components/CustomSelect.astro
```diff
@@ -350,14 +357,29 @@ const hasIcon = Astro.slots.has('icon');
...
+          Math.abs(window.scrollX - openScrollX) <= 2 &&
+          Math.abs(window.scrollY - openScrollY) <= 2
+        ) {
+          return;
+        }
```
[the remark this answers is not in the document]
IEvangelist: Extracted the tolerance into a named `scrollSettleTolerancePx` constant (declared once at module scope, with a comment explaining the sub-pixel rounding / high-DPI mobile rationale) and referenced it from the scroll handler. Pushed in d1c3b2e8.

PR #1491, opened 2026-08-13
ci: auto-resolve data/content conflicts in release-branch merge

IEvangelist:
## Problem

The **Update Release Branch** workflow (`update-release-branch.yml`) merges `main` into the active release branch (currently `release/13.5`) on every push to `main`. It has been failing on essentially every run — e.g. [run 31735187914](https://github.com/microsoft/aspire.dev/actions/runs/31735187914) — with `Merge conflict detected ... Manual intervention required`.

`main` and the release branch each accumulate **independent, non-fast-forwardable** changes to the same files, so a plain `git merge` always conflicts:

- **Generated data** under `src/frontend/src/data/**` — `aspire-integrations.json`, `github-stats.json`, `twoslash/aspire.d.ts`, and the **version-stamped** `pkgs/**` and `ts-modules/**` files. Because each side renames the same base file to a different version (e.g. `...260730.1.json` vs `...260804.1.json`), git reports **rename/rename** conflicts. (72 files in the latest run.)
- **Curated site content** under `src/frontend/src/content/**` — e.g. the release-specific `✨ Aspire 13.5 is available!` banner and the localized `index.mdx` pages. (19 files in the latest run.)

A blanket `-X theirs` would regress the live site (revert the 13.5 banner back to 13.4); a blanket `-X ours` would freeze the generated data. Neither is correct.

## Fix

Replace the hard-fail with **path-scoped auto-resolution** in the merge step:

| Path | Resolution | Why |
|---|---|---|
| `src/frontend/src/data/**` + `src/frontend/src/assets/samples/**` | take **main** (rewrite subtree to match) | Regenerated from source-of-truth (NuGet + GitHub) by `update-integration-data`, which only ever targets `main`. Rewriting the whole subtree also clears rename-orphaned version-stamped files. |
| `src/frontend/src/content/**` | keep **release** on conflict | The release branch owns its published/localized content; non-conflicting content from `main` still merges in normally. |
| anything else | **fail** (manual intervention) | Safety valve for genuinely unexpected conflicts. |

The clean-merge fast path is unchanged.

## Validation

Simulated the exact algorithm against `upstream/release/13.5` + `upstream/main` (the failing run's commit):

- 91 conflicts → **0** unresolved; no leftover conflict markers.
- `src/frontend/src/data` ends **byte-identical to main**; the orphaned `...260730.1.json` package files are removed.
- The `✨ Aspire 13.5 is available!` banner is preserved in `index.mdx` / `community/index.mdx`.
- `bash -n` and YAML parse both pass.

### DIFF
--- a/.github/workflows/update-release-branch.yml
+++ b/.github/workflows/update-release-branch.yml
@@ -75,14 +75,139 @@ jobs:
 
       - name: Merge main into release branch
         if: ${{ steps.find-branch.outputs.branch != '' }}
+        env:
+          BRANCH: ${{ steps.find-branch.outputs.branch }}
         shell: bash
         run: |
+          set -euo pipefail
+
           git fetch origin main
-          if git merge origin/main --no-edit; then
-            echo "Merged main into ${{ steps.find-branch.outputs.branch }} successfully."
-            git push
-          else
-            echo "::error::Merge conflict detected between main and ${{ steps.find-branch.outputs.branch }}. Manual intervention required."
+          MAIN_SHA="$(git rev-parse origin/main)"
+
+          # The release branch is a curated snapshot of the shipped site. Merging
+          # main into it conflicts in three, and only three, well-understood ways:
+          #   1. Generated data          -> main is authoritative (mirror it).
+          #   2. Curated content/imagery -> the release branch is authoritative.
+          #   3. Anything else           -> a human must decide; abort cleanly.
+          #
+          # (1) is the exact output set of the update-integration-data workflow
+          # (its `allowed-files`); those paths are 100% machine-generated, so main
+          # always wins and mirroring also clears the version-stamped rename/rename
+          # orphans they routinely produce. (2) is site content plus documentation
+          # imagery, which the release branch pins to release-specific versions (the
+          # "Aspire 13.5 is available" banner, 13.5 dashboard screenshots, ...) and
+          # must never regress. (3) is a genuine divergence we refuse to guess at.
+
+          GEN_FILES="
+          src/frontend/src/data/aspire-integrations.json
+          src/frontend/src/data/github-stats.json
+          src/frontend/src/data/samples.json
+          src/frontend/src/data/twoslash/aspire.d.ts
+          "
+          GEN_DIRS="
+          src/frontend/src/data/pkgs
+          src/frontend/src/data/ts-modules
+          src/frontend/src/assets/samples
+          "
+
+          is_generated() {
+            case "$1" in
+              src/frontend/src/data/aspire-integrations.json|\
+              src/frontend/src/data/github-stats.json|\
+              src/frontend/src/data/samples.json|\
+              src/frontend/src/data/twoslash/aspire.d.ts|\
+              src/frontend/src/data/pkgs/*|\
+              src/frontend/src/data/ts-modules/*|\
+              src/frontend/src/assets/samples/*) return 0 ;;
+              *) return 1 ;;
+            esac
+          }
+
+          # assets/samples/** is generated and is matched by is_generated first;
+          # it is excluded here defensively in case the checks are reordered.
+          is_curated() {
+            case "$1" in
+              src/frontend/src/assets/samples/*) return 1 ;;
+              src/frontend/src/content/*) return 0 ;;
+              src/frontend/src/assets/*) return 0 ;;
+              *) return 1 ;;
+            esac
+          }
+
+          # Keep the release branch's version, honoring a release-side deletion.
+          keep_release() {
+            if git cat-file -e "HEAD:$1" 2>/dev/null; then
+              git checkout HEAD -- "$1"
+              git add -- "$1"
+            else
+              git rm -q --force --ignore-unmatch -- "$1" >/dev/null
+            fi
+          }
+
+          # A generated path needs syncing when it has a conflict stage or its
+          # merged content differs from main.
+          needs_mirror() {
+            [ -n "$(git ls-files -u -- "$1")" ] && return 0
+            git diff --quiet "$MAIN_SHA" -- "$1" 2>/dev/null || return 0
+            return 1
+          }
+
+          git merge --no-commit --no-ff origin/main || true
+
+          if ! git rev-parse -q --verify MERGE_HEAD >/dev/null; then
+            echo "Already up to date with origin/main; nothing to merge."
+            exit 0
+          fi
+
+          # Triage conflicts: generated ones are handled by the mirror step below;
+          # keep-release for curated paths; flag anything else as unresolvable.
+          unresolved=0
+          while IFS= read -r f; do
+            [ -z "$f" ] && continue
+            if is_generated "$f"; then
+              continue
+            elif is_curated "$f"; then
+              keep_release "$f"
+            else
+              echo "::error::Unresolvable merge conflict in ${f} (not a generated or curated path)."
+              unresolved=1
+            fi
+          done < <(git diff --name-only --diff-filter=U)
+
+          if [ "$unresolved" -ne 0 ]; then
             git merge --abort
+            echo "::error::Automated merge of main into ${BRANCH} aborted; a human must resolve the conflicts above."
             exit 1
           fi
+
+          # Force every generated producer path to match main exactly. This clears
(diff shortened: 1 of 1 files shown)


.github/workflows/update-release-branch.yml
```diff
@@ -76,13 +76,62 @@ jobs:
...
+          # Generated data and sample assets are regenerated from source-of-truth
+          # data on main (NuGet + GitHub) by the update-integration-data workflow,
+          # which only ever targets main. Take main's versions wholesale so the
+          # release branch always carries the freshest data. Replacing the whole
+          # subtree (rm + checkout) also clears rename-orphaned versioned files
+          # under pkgs/ and ts-modules/ that would otherwise linger as duplicates.
+          git rm -r --force --quiet --ignore-unmatch \
+            src/frontend/src/data \
+            src/frontend/src/assets/samples
+          git checkout "$MAIN_SHA" -- \
```
adamint: Can we scope this to the generated files that actually conflicted? One conflict here replaces the entire `src/frontend/src/data` subtree from main, including tracked schema, testimonials, sessions, and integration files this workflow does not regenerate. Any clean release-only edits in those files are silently overwritten and recorded as part of the merge resolution. I think this should resolve from the unmerged-file list and limit the take-main set to the producer workflow outputs.

.github/workflows/update-release-branch.yml
```diff
@@ -76,13 +76,62 @@ jobs:
...
+          # Curated site content under content/ is owned by the release branch
+          # (for example the release-specific "Aspire 13.5 is available" banner and
+          # the localized index pages). Keep the release version when it conflicts
+          # with main; non-conflicting content changes from main still merge in.
+          unresolved=0
+          while IFS= read -r file; do
+            [ -z "${file}" ] && continue
+            case "${file}" in
+              src/frontend/src/content/*)
+                git checkout --ours -- "${file}"
```
adamint: This exits immediately under `set -e` for a modify/delete conflict where the release branch deleted the path: `git checkout --ours -- "$file"` reports that the path has no ours version. That skips the later `git merge --abort` and actionable error. Could we handle the missing-ours case with `git rm` (or resolve by conflict status) instead?

.github/workflows/update-release-branch.yml
```diff
@@ -76,13 +76,62 @@ jobs:
...
+          # which only ever targets main. Take main's versions wholesale so the
+          # release branch always carries the freshest data. Replacing the whole
+          # subtree (rm + checkout) also clears rename-orphaned versioned files
+          # under pkgs/ and ts-modules/ that would otherwise linger as duplicates.
+          git rm -r --force --quiet --ignore-unmatch \
+            src/frontend/src/data \
+            src/frontend/src/assets/samples
+          git checkout "$MAIN_SHA" -- \
+            src/frontend/src/data \
+            src/frontend/src/assets/samples
```
adamint: This also exits under `set -e` if either subtree is absent on main (for example, main deletes or renames `assets/samples`): `git checkout "$MAIN_SHA" -- <path>` has no `--ignore-unmatch`, after the preceding `git rm` has already changed the index. Could we guard each checkout with `git cat-file -e "$MAIN_SHA:<path>"` and route failures through the abort branch?

adamint (approved):
The resolver still fails safely for authored-code conflicts, and the exact workflow shell was exercised in throwaway git repositories. I left three must-fix comments: a single conflict resets unrelated release-owned data, release-side deletions fail before the actionable abort path, and a main-side subtree deletion fails after staging broad removals. These are fixable in place, so approving with the comments.

.github/workflows/update-release-branch.yml
IEvangelist: Good catch — fixed. The resolver no longer checks out the whole `src/frontend/src/data` subtree. It now iterates the unmerged-file list (`git diff --name-only --diff-filter=U`) and the take-main set is limited to the exact producer `allowed-files` from `update-integration-data`: `aspire-integrations.json`, `github-stats.json`, `samples.json`, `twoslash/aspire.d.ts`, `pkgs/**`, `ts-modules/**`, and `assets/samples/**` (see `is_generated`).

The hand-authored files under `data/` — `schemas/**`, `testimonials.json`, `aspireconf-sessions.ts`, `integration-docs.json`, `aspire-integration-names.json`, `thanks-license-titles.ts` — are never in that set, so a conflict in any of them falls through to the `else` arm and aborts for a human instead of being silently overwritten.

I validated this against the real `release/13.5` ← `main` merge: the generated paths end byte-identical to main, while `testimonials.json`, `aspireconf-sessions.ts`, `integration-docs.json`, and `aspire-integration-names.json` are preserved at their release-branch versions.

.github/workflows/update-release-branch.yml
IEvangelist: Fixed. There's no longer a `git checkout --ours` in the resolver. Curated paths are resolved by a `keep_release` helper that first checks whether the path still exists on `HEAD`:

```bash
keep_release() {
  if git cat-file -e "HEAD:$1" 2>/dev/null; then
    git checkout HEAD -- "$1"
    git add -- "$1"
  else
    git rm -q --force --ignore-unmatch -- "$1" >/dev/null
  fi
}
```

For a modify/delete conflict where the release branch deleted the path, there is no `HEAD` blob, so it takes the `git rm --ignore-unmatch` branch and stages the deletion instead of exiting under `set -e`. I added a regression test for exactly this shape (content modify/delete where the release side deleted the file) — it stays deleted and the merge still completes with exit 0.

.github/workflows/update-release-branch.yml
IEvangelist: Fixed. Every take-main checkout is now guarded by an existence check on `main`, and the removal paths use `--ignore-unmatch`:

```bash
for p in $GEN_FILES $GEN_DIRS; do
  if git cat-file -e "$MAIN_SHA:$p" 2>/dev/null; then
    if needs_mirror "$p"; then
      git rm -r -q --force --ignore-unmatch -- "$p" >/dev/null 2>&1 || true
      git checkout "$MAIN_SHA" -- "$p"
      git add -A -- "$p"
    fi
  else
    # main deleted/renamed the path -> drop it from the release branch too
    git rm -r -q --force --ignore-unmatch -- "$p" >/dev/null 2>&1 || true
  fi
done
```

The `git checkout "$MAIN_SHA" -- "$p"` only runs inside the branch where `git cat-file -e "$MAIN_SHA:$p"` already confirmed the path exists on main, so it can no longer hit the unmatched-path error after the preceding `git rm` changed the index. If main deleted or renamed the subtree, the guard routes it to the `else` branch (remove locally). I validated the whole flow against the real `release/13.5` ← `main` merge (which includes `pkgs/**` and `ts-modules/**` rename/rename churn) and it completes cleanly with the generated tree byte-identical to main.

.github/workflows/update-release-branch.yml
```diff
@@ -76,13 +76,62 @@ jobs:
...
+            case "${file}" in
+              src/frontend/src/content/*)
+                git checkout --ours -- "${file}"
+                git add -- "${file}"
+                ;;
+              *)
+                echo "::error::Merge conflict in ${file} is outside the auto-resolvable paths (data, assets, content)."
+                unresolved=1
+                ;;
+            esac
```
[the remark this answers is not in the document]
IEvangelist: Thanks for flagging — I checked this carefully and the nested paths are actually matched, so no false abort occurs. In a POSIX `case` statement, `*` is a glob wildcard that spans `/` (unlike filename expansion in the shell, `case` pattern matching gives `/` no special treatment), so `src/frontend/src/content/*` matches `src/frontend/src/content/docs/community/index.mdx` and every other nested file.

Quick demonstration:

```bash
$ f=src/frontend/src/content/docs/community/index.mdx
$ case "$f" in src/frontend/src/content/*) echo match ;; *) echo no ;; esac
match
```

I also confirmed it empirically against the real `release/13.5` ← `main` merge: all 19 conflicting `content/docs/**` files — including deeply nested ones like `community/index.mdx`, `ja/docs.mdx`, and the localized `*/index.mdx` variants — are matched by the content arm (now named `is_content`) and resolved to the release branch's version. None fell through to the abort arm.

PR #1534, opened 2026-08-18
Clarify ASPIRE011 run preflight wording

IEvangelist:
## Summary

- clarify that missing `dnx` produces `ASPIRE011` during run preflight
- distinguish launch-command preparation from a regular build

## Validation

- `git diff --check`
- one file with one insertion and one deletion on current `main`
- repository Prettier reports only two pre-existing deltas in this file; formatting a temporary copy does not modify the changed sentence


alistairmatthews (approved):
Looks good. Approved.

PR #1537, opened 2026-08-19
Render theme-aware sample images that link to their full-size asset

IEvangelist:
## Summary

Theme-aware sample screenshots stopped rendering on generated sample reference pages — for example [`/reference/samples/image-gallery/`](https://aspire.dev/reference/samples/image-gallery/) showed broken image icons instead of the light/dark previews.

Sample READMEs frequently link a screenshot thumbnail to its full-size asset with the ubiquitous GitHub pattern `[alt](full)`. That parses to a paragraph of `link` nodes wrapping `image` nodes rather than bare `image` nodes, so `getStandaloneImageNodes` (which only inspected direct `image` children) stopped recognizing them as standalone sample images. The blocks then fell through to raw HTML rendering and surfaced as broken `` tags — the `~/assets/...` alias only resolves through Astro's build-time image importer, never in the browser — instead of being routed to the theme-aware sample media pipeline.

This change unwraps `link` children whose sole content is a single image, so both the light/dark theme-pair detection and the standalone-block skip logic see the inner images again. Bare images keep their existing behavior.

None.

## Validation

- Added `src/frontend/tests/unit/sample-readme-images.vitest.test.ts`, which parses the real README markdown via `satteri` (the same parser `SampleDetail.astro` uses) and asserts the linked light/dark pair is detected as a standalone theme block. The new cases fail on `main` and pass with this change.
- `pnpm test:unit` — all 31 files / 376 tests pass.
- ESLint clean on the changed files; Prettier-formatted.

### DIFF
--- a/src/frontend/src/utils/sample-readme-images.ts
+++ b/src/frontend/src/utils/sample-readme-images.ts
@@ -10,22 +10,65 @@ function isWhitespacePhrasing(node: PhrasingContent): boolean {
   return node.type === 'text' && node.value.trim() === '';
 }
 
+/**
+ * Resolve a paragraph child down to the standalone image it represents, or
+ * `null` when it isn't one.
+ *
+ * A bare `image` node is returned as-is. GitHub READMEs also very commonly link
+ * a screenshot thumbnail to its full-size asset — `[![alt](thumb)](full)` —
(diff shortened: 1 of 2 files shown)


alistairmatthews (approved):
Theme-aware sample images now render correctly in my tests. Approved.

PR #1544, opened 2026-08-19
Reword landing hero: 'Open source and free' and 'Compose' verb

IEvangelist:
## What & why

Reword the landing-page hero based on feedback:

- **Eyebrow:** "Open source and developer first" → **"Free and open source"** across English and all 14 localized `i18n` files.
- **Title verb:** "Model distributed apps in code." → **"Compose distributed apps in code."** across all 15 locales. Feedback was that the "Model" verb reads poorly / is hard to understand at first.

Per-locale verbs were taken from each locale's existing hero `summary` string (which already used that locale's word for "composing"), so the translations stay consistent with copy already on the page.

## Accent markup

The purple accent on "code." previously used a `<strong>` wrapper. Since the weight is handled by CSS, the wrapper is now a `<span class="accent">` and the color is applied in `HomeHero.astro` via `:global(.accent)` — matching the existing `set:html` convention used by `SampleCard.astro` and `Banner.astro`. (The prior `.home-hero-copy h1 strong` rule lacked `:global`, so it wasn't reliably applying to the `set:html`-injected title.)

## Also updated

- `index.mdx` SEO `<title>` and Starlight frontmatter `hero.title`.
- Unit (`custom-components.vitest.test.ts`) and e2e (`homepage.spec.ts`) expectations for the new title.

## Review fixes

- **Korean heading punctuation:** moved the sentence-final period out of the accent span so the accessible heading reads as one sentence — `분산 앱을 <span class="accent">코드로</span> 구성하세요.`
- **e2e accent coverage:** `homepage.spec.ts` now asserts the `h1 > span.accent` markup (presence + text) and its computed `color`, resolving `--home-purple` at runtime so the check stays theme-independent and guards against markup/color regressions.

## Intentionally left unchanged

- CapabilityGrid test fixtures containing "Model distributed apps" (unrelated feature).
- The separate "Model the app in code" step-card title (different element from the hero).


src/frontend/src/content/i18n/ko.json
```diff
@@ -188,8 +188,8 @@
   },
   "home": {
     "hero": {
-      "eyebrow": "오픈 소스, 개발자 우선",
-      "title": "분산 앱을 <strong>코드로.</strong> 모델링하세요",
+      "eyebrow": "오픈 소스이며 무료",
+      "title": "분산 앱을 <span class=\"accent\">코드로.</span> 구성하세요",
```
adamint: `코드로.` now ends a sentence inside the accent span, but `구성하세요` follows it, so the accessible heading reads “분산 앱을 코드로. 구성하세요.” Move the period to the end: `분산 앱을 <span class="accent">코드로</span> 구성하세요.`

src/frontend/tests/e2e/homepage.spec.ts
```diff
@@ -11,7 +11,7 @@ test.beforeEach(async ({ page }) => {
 test('renders a complete semantic landing page without horizontal overflow', async ({ page }) => {
   await expect(page.locator('main h1')).toHaveCount(1);
   await expect(
-    page.getByRole('heading', { level: 1, name: 'Model distributed apps in code.' })
+    page.getByRole('heading', { level: 1, name: 'Compose distributed apps in code.' })
```
adamint, line 14: This proves only the flattened heading text. The behavior changed from `<strong>` to `span.accent` and now relies on a new scoped/global CSS selector; either the markup or color can regress while this still passes. Add an assertion for `h1 > span.accent` and its computed `color`.

adamint (approved):
The copy and selector change look good overall. Two fixes before merge: the Korean heading has sentence punctuation in the middle, and the e2e coverage should assert the new accent markup and color rather than only flattened text.

src/frontend/src/content/i18n/ko.json
IEvangelist: Good catch. Moved the period out of the accent span to the end of the sentence, exactly as suggested: `분산 앱을 <span class="accent">코드로</span> 구성하세요.` The accessible heading now reads as a single sentence.

src/frontend/tests/e2e/homepage.spec.ts
IEvangelist: Added assertions right after the heading check: `main h1 > span.accent` must exist (count 1) and read `code.`, and its computed `color` must resolve to `--home-purple` and differ from the heading text color. The purple is resolved at runtime via a probe element so it stays theme-independent. Verified passing on desktop, tablet, and mobile Chromium.

PR #1545, opened 2026-08-19
Add ContainerImages component and surface container images across hosting docs

IEvangelist:
## What

Adds a new `ContainerImages` Astro component that renders the container images each Aspire hosting integration runs, and embeds it across every hosting/emulator article. Also adds a standalone **Container images reference page** that catalogs every image in one place.

The data comes from a new `src/frontend/src/data/container-images.json`, generated (via the `container-images` skill) from the `*ContainerImageTags.cs` files in `microsoft/aspire` at `main`.

## Component

Each image renders as an on-theme card showing:

- the friendly image name + a registry glyph (muted for companion/tooling images);
- a **Companion** chip linking each tooling image (pgAdmin, Kafka UI, Redis Commander, …) to its primary;
- the full `registry/image:tag` as a **copyable** reference (rendered via the site's Expressive Code block, so it gets a native copy button and wraps only at `/` path boundaries, never mid-token);
- **Also available** chips for alternate tags (e.g. Cosmos `vnext-latest`, RabbitMQ `4.3-management`);
- a **Source** link to the exact `*ContainerImageTags.cs`.

It uses Starlight CSS variables (adapts to light/dark automatically) and is responsive — multiple columns on wide viewports, a single column in the docs content area.

Props: `package` (required), optional `only` (filter to specific image names), optional `title`.

## Container images reference page

New `/reference/container-images/` page (`ContainerImageCatalog` component) that enumerates **every** hosting container image in one place, grouped by category (databases, caching, messaging, observability, security, reverse proxies, Azure emulators). It renders as a responsive 2-column card grid mirroring the integration gallery — each card shows the copyable image reference, its registry, any companion relationship, and a docs link to the introducing integration. Categories are derived automatically from each integration's docs href, so newly added integrations are grouped without a hand-maintained map. Added to the sidebar under **Aspire Reference** with the standard 15-language translations.

## Rollout

Embedded into all **24 hosting packages across 26 host/emulator articles** — placed where each container is introduced (primary intro for databases/caching/messaging; the `RunAsEmulator` section for Azure emulators; the emulator subsection for Keycloak/YARP/Storage). Azure Storage's Azurite emulator appears in the blobs/queues/tables articles.

## Versioning

Pinned Aspire package versions in the wired articles now use the shared `%ASPIRE_VERSION%` placeholder that flows through docs generation, instead of hard-coded version strings, so they stay current automatically.

## Docs accuracy

While wiring the cards, I corrected three stale prose image references the cards surfaced:

- **Redis Commander**: `docker.io/rediscommander/redis-commander` -> `ghcr.io/joeferner/redis-commander`
- **Cosmos DB emulator**: `mcr.microsoft.com/cosmosdb/emulator` -> `.../cosmosdb/linux/azure-cosmos-emulator`
- **Event Hubs emulator**: dropped the malformed `.../eventhubs-emulator/latest`

## Tests & validation

- Added render fixtures + a prop contract for the component, asserting the Expressive Code copy button (`title="Copy to clipboard"`, image reference wired via `data-code`).
- `pnpm test:unit` passes (388/388); ESLint clean; `check-data` passes; light + dark verified for primary + companions, alternate tags, long-reference wrapping, copy button, and WCAG AA contrast on the reference page (companion/count chips).


src/frontend/src/content/docs/integrations/databases/postgres/postgres-host.mdx
```diff
@@ -122,6 +123,8 @@ await builder.addNodeApp("api", "./api", "index.js")
     When you reference a PostgreSQL resource from the AppHost, Aspire makes several properties available to the consuming project, such as connection URIs, hostnames, and port numbers. For a complete list of these properties and per-language connection examples, see [Connect to PostgreSQL](../postgres-connect/).
 </Aside>
 
+<ContainerImages package="Aspire.Hosting.PostgreSQL" />
```
adamint: This page installs `Aspire.Hosting.PostgreSQL` 13.3.0 (line 67), and later explicitly says 13.3 uses PostgreSQL 17 (line 408), but this card resolves to `postgres:18.3` from data generated against `main`. A reader following the page will copy an image that does not match the documented package, and these major versions have incompatible data layouts. Can we version the image data against the package shown by the docs, or update the package examples in the same change?

src/frontend/src/components/ContainerImages.astro
```diff
@@ -0,0 +1,351 @@
...
+    const button = target?.closest<HTMLButtonElement>('button[data-ci-copy]');
+    if (!button) return;
+
+    try {
+      await navigator.clipboard.writeText(button.dataset.ciCopy ?? '');
+    } catch {
+      return;
+    }
+
+    button.classList.add('is-copied');
```
adamint: This only changes the icon, and both icons are `aria-hidden`, so the button's accessible name remains “Copy …” after success. Clipboard failures are also silent. Screen-reader users never hear confirmation, and anyone with a denied clipboard gets no feedback. Can we expose success and failure through an `aria-live` status and cover the interaction?

adamint (dismissed):
The PostgreSQL card currently shows an image from a different Aspire release than the page installs, and the copy interaction needs programmatic success/failure feedback. I left both issues inline.

src/frontend/src/components/ContainerImages.astro
```diff
@@ -0,0 +1,351 @@
...
+    window.setTimeout(() => button.classList.remove('is-copied'), 1500);
+  });
+</script>
```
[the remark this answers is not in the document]
IEvangelist: Good catch — this is addressed in the redesign. The custom copy button and its document-level clipboard handler are gone; the image block now renders through Expressive Code's `<Code>`, which ships its own single copy button. The only remaining document-level listener is the language-pivot script, and it's guarded by a global `window.__aspireLangPivot` sentinel so it registers exactly once per document, no matter how many cards are on the page.

src/frontend/src/content/docs/integrations/cloud/azure/azure-cosmos-db/azure-cosmos-db-host.mdx
```diff
@@ -143,7 +144,9 @@ await cosmos.runAsEmulator();
 </TabItem>
 </Tabs>
 
-When you call `RunAsEmulator`, it configures the Cosmos DB resource to run locally using the [Azure Cosmos DB Emulator](https://learn.microsoft.com/azure/cosmos-db/local-emulator). The emulator is accessible to Aspire as a container image (`mcr.microsoft.com/cosmosdb/emulator`). Aspire creates and starts the container when the AppHost starts.
+When you call `RunAsEmulator`, it configures the Cosmos DB resource to run locally using the [Azure Cosmos DB Emulator](https://learn.microsoft.com/azure/cosmos-db/local-emulator). The emulator is accessible to Aspire as a container image (`mcr.microsoft.com/cosmosdb/linux/azure-cosmos-emulator`). Aspire creates and starts the container when the AppHost starts.
+
+<ContainerImages package="Aspire.Hosting.Azure.CosmosDB" />
```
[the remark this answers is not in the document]
IEvangelist: Fixed. There are no remaining `mcr.microsoft.com/cosmosdb/emulator` references under docs — both files you flagged now use `mcr.microsoft.com/cosmosdb/linux/azure-cosmos-emulator` (`azure-cosmos-db-get-started.mdx:26` and `efcore/azure-cosmos-db/azure-cosmos-db-host.mdx:191`), matching the hosting page and the ContainerImages card.

src/frontend/src/content/docs/integrations/cloud/azure/azure-event-hubs/azure-event-hubs-host.mdx
```diff
@@ -169,7 +170,7 @@ The `AddConsumerGroup` call registers a consumer group named `messagesConsumer`
 
 ## Run as emulator
 
-The hosting integration supports running Event Hubs locally using the `mcr.microsoft.com/azure-messaging/eventhubs-emulator/latest` container image. This lets you develop and test without an Azure subscription or an existing Event Hubs namespace.
+The hosting integration supports running Event Hubs locally using the `mcr.microsoft.com/azure-messaging/eventhubs-emulator` container image. This lets you develop and test without an Azure subscription or an existing Event Hubs namespace.
```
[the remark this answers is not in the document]
IEvangelist: Fixed. The malformed `.../eventhubs-emulator/latest` reference is gone — `azure-event-hubs-get-started.mdx:26` now uses `mcr.microsoft.com/azure-messaging/eventhubs-emulator`, consistent with the hosting article and the card.

src/frontend/src/content/docs/integrations/databases/postgres/postgres-host.mdx
IEvangelist: You're right that the card can show a newer tag than the pinned package. That's by design: the catalog is generated against microsoft/aspire@main so it always reflects the current defaults, and pinning each card to the version in a given doc's install snippet isn't tractable across the site (those snippets pin a range of versions). Rather than let the drift be silent, I've added a provenance note under every card stating that the tags reflect the latest defaults on the aspire **main** branch and may be newer than the version pinned by the package you install. The Postgres 17-vs-18 default change is also already documented on this page (the "13.4 bumped the default to 18.3" section, including the stay-on-17 option), so a reader has the full picture.

src/frontend/src/components/ContainerImages.astro
IEvangelist: Moot now — the custom copy button was removed in the redesign. The image block renders through Expressive Code's `<Code>`, which provides its own accessible copy button with built-in success/failure feedback, so the aria-hidden icon and silent-failure concerns no longer apply.

IEvangelist:
Extended scope in `384bbb91`: replaced hard-coded Aspire package versions in the English docs with the existing `%ASPIRE_VERSION%` placeholder so they flow through docs gen (same convention already used in `seed-database.mdx` and `add-aspire-existing-app.mdx`). 51 files, 56 lines — mostly `aspire.config.json` install pins, plus the `PackageReference`/SDK forms in the testing and ASPIRE008 pages.

Deliberately left untouched:
- Intentional wildcards (`@*`, `Version="*"`) in the C# tabs.
- `whats-new/*` — version-anchored history.
- `diagnostics/aspire007.mdx` — its `9.0.0`/`9.2.0` example documents the SDK version floor the diagnostic is about.
- Translated copies (`/ja/`, etc.) — regenerated by the translation pipeline, so they self-heal from the English source.

alistairmatthews (approved):
Container image cards work well. Rendered links work. Container images catalog renders correctly. Approved.

PR #1542, opened 2026-08-19
Add a per-release Community section to What's new pages

IEvangelist:
## What & why

Gives credit where it's due. Adds an [Astro-style](https://astro.build/blog/astro-710/#community) **Community** section to every "What's new" release page: a running **core-team roster** plus the **full list of community contributors** credited for that specific release.

Previously each release doc had a hand-written `🙏 Community contributions` paragraph that shouted out only a curated handful of people; `/community/contributors/` shows all-time contributors per repo but nothing per release.

## What's included

**New reusable component** — `src/frontend/src/components/ReleaseCommunity.astro`
- Renders the core-team roster (avatar + name + profile link) and the per-release _"Special thanks to everyone who contributed to Aspire X.Y … including:"_ list, followed by the standard contributing call-to-action.
- Used as `<ReleaseCommunity version="13.5" />`.

**Core-team roster** — `src/frontend/src/data/core-team.ts`
- Hand-maintained running list (handle + name + optional social link). Avatars derive from `https://github.com/<handle>.png`.
- ⚠️ **Seeded with a clearly-marked placeholder list — please confirm/adjust the exact roster before merge.**

**Generated per-release data** — `src/frontend/src/data/release-contributors.json` + `scripts/update-release-contributors.ts`
- Pulls contributors from GitHub's release-notes API for each tag range (`v8.2.2..v9.0.0` … `v13.4.0..v13.5.0`), collecting both PR authors and co-authors.
- Filters out bots and current core-team members (so they're credited once, in the roster — matching Astro).
- Wired as `pnpm update:release-contributors` (mirrors the existing `update:github-stats` pattern; run per release to refresh).

**Wiring** — all **12 English release docs** (9.0 → 13.5)
- Docs with existing curated shout-outs keep them as the intro; the roster + full list render below.
- Docs without get a standardized `🙏 Community` section.

**Test** — added a render case for the component to `custom-components.vitest.test.ts`.

## Notes for reviewers

- **Core-team roster is a placeholder** — the seeded handles need your confirmation (names, exact membership, preferred social links).
- Japanese `ja/` release docs are intentionally out of scope here; they'll pick this up through the normal Lunaria translation sync.
- Contributor lists reflect merged-PR authorship per tag range (code/PR contributors); reviewers/testers aren't captured by the API.

## Validation

- `pnpm build:skip-search` compiles all pages and passes link validation (the only failure was an unrelated `EBUSY` file-lock during `dist` asset rename on Windows).
[...]

PR #1548, opened 2026-08-19
Order core team roster alphabetically by name

IEvangelist:
## What & why

Follow-up to #1542. Orders the hand-maintained core-team roster in `core-team.ts` alphabetically by **display name** so the Community section renders team members in a predictable order.

## Changes

- Reordered the `coreTeam` array alphabetically by `name` (first-name-first). No members added or removed — 14 before and after.
- Added a note to the doc comment that the roster is ordered by display name, so future additions preserve the convention.

## Notes

- Interpreted "order by name" as alphabetical by the display `name` field as written (e.g. the three `David …` entries stay grouped: Fowler → Negstad → Pine). Happy to switch to last-name ordering if preferred.
- `handle` / avatar / social-link behavior is unchanged; `coreTeamHandles` is derived from the array so de-duplication is unaffected by order.

### DIFF
--- a/src/frontend/src/data/core-team.ts
+++ b/src/frontend/src/data/core-team.ts
@@ -8,9 +8,10 @@
  * maintained by hand.
  *
  * Members listed here are intentionally excluded from the generated per-release
- * "Special thanks" list so they are credited once, in the roster. Keep the
- * roster current as membership changes, and set a member's `social` link to
- * override the default GitHub profile link.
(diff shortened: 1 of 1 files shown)


mitchdenny (approved):
Reviewed the roster reorder at 4ba7d313ed9e42d2432a9bb095ed789850589ae7. The same 14 members are preserved, the display-name ordering is correct, and the affected unit tests pass.

PR #1561, opened 2026-08-20
Update dashboard telemetry limit documentation

JamesNK:
## Summary

- Clarify how dashboard telemetry limits evict, truncate, or reject telemetry after limits are reached.
- Document that `Dashboard:TelemetryLimits:MaxMetricsCount` applies per dimension.
- Align the configuration reference with microsoft/aspire#19525 commits `be8131cfc8a40fbfe837c59e8dd0633250d18793` and `77e42f4a87fbe17939ebdb496903b1d45645d492`.

None.

## Validation

- `pnpm test:unit:docs`


src/frontend/src/content/docs/dashboard/configuration.mdx
```diff
@@ -282,18 +282,18 @@ The resource service client authentication is configured with `Dashboard:Resourc
 
 ### Telemetry limits
 
-Telemetry is stored in memory. To avoid excessive memory usage, the dashboard has limits on the count and size of stored telemetry. When a count limit is reached, new telemetry is added, and the oldest telemetry is removed. When a size limit is reached, data is truncated to the limit.
+Telemetry is stored in memory. To avoid excessive memory usage, the dashboard limits stored telemetry. Log, trace, and metric retention limits evict the oldest stored values when full; attribute and span-event limits truncate incoming data, and the resource limit rejects telemetry for new resources after the limit is reached.
```
adamint, line 286: Could we mention that this is a soft cap? `GetOrAddResource` does a non-atomic count check, so concurrent new resources can exceed `MaxResourceCount`. As written, “rejects ... after the limit is reached” reads like a hard ceiling. We should update the Dashboard README too so the two references stay aligned.

IEvangelist (approved):

PR #1666, opened 2026-09-15
Fix release branch version conflict handling

IEvangelist:
## Summary

- resolve `src/frontend/config/aspire-versions.mjs` by taking main's config structure while retaining version assignments already owned by the release branch
- preserve newly added exports and placeholders such as `currentAspirePreviewVersion` and `%ASPIRE_VERSION_PREVIEW%`
- preserve a release-owned preview assignment when present; fall back to main only when the older release file has no such assignment
- continue keeping curated documentation content and imagery from the release branch
- abort explicitly if either config shape no longer contains the expected version assignments

## Semantic consumers verified

The simulated merged tree contains functional preview-version consumers in:

- `src/frontend/src/content/docs/integrations/cloud/azure/azure-ai-foundry/azure-ai-foundry-host.mdx`
- `src/frontend/src/content/docs/integrations/cloud/azure/azure-data-explorer.mdx`
- `src/frontend/src/content/docs/integrations/devtools/browser-logs.mdx`
- `src/frontend/src/content/docs/reference/cli/commands/aspire-doctor.mdx`
- `src/frontend/tests/unit/aspire-version-placeholders.vitest.test.ts`

## Release pin disposition

The incoming unit test intentionally asserts that the stable and preview config constants match the site AppHost SDK and FrontDoor package. Repository release guidance requires the same alignment and classifies stale or nonexistent values as critical.

`release/13.6` currently has `currentAspireVersion = 13.6.0`, while its AppHost remains on stable `13.5.3` and FrontDoor `13.5.3-preview.1.26425.3`. As of September 15, 2026, public NuGet lists no 13.6 AppHost SDK or FrontDoor preview package. Therefore the branch is in a pre-release inconsistent state; this PR does not establish or change those pins, and should not by itself be treated as evidence that release/13.6 is ready to sync.

The smallest final policy-compliant release fix is to update the AppHost SDK, FrontDoor package, stable config assignment, and preview config assignment together once matching published 13.6 versions exist. If the branch must be made internally consistent before then, the release owner must explicitly choose whether to retain the currently published 13.5.3 pair instead of advertising 13.6 package versions.

## Validation

- parsed and formatting-checked the workflow with Prettier
- extracted the full merge step and passed `bash -n`
- confirmed `git merge-tree --write-tree upstream/release/13.6 upstream/main` reports exactly the version config and two dashboard documentation conflicts
- ran the updated workflow logic in detached temporary worktrees without mutating `release/13.6`
- verified the current older release file falls back to main's newly introduced preview assignment
- verified a synthetic future release-owned preview assignment is preserved instead of overwritten
- imported the resolved config and verified the preview placeholder map remains functional


.github/workflows/update-release-branch.yml
```diff
@@ -164,7 +232,11 @@ jobs:
           unresolved=0
           while IFS= read -r f; do
             [ -z "$f" ] && continue
-            if is_generated "$f"; then
+            if [ "$f" = "src/frontend/config/aspire-versions.mjs" ]; then
```
eerhardt, line 235: How often do you think this file is going to change? Is it worth this?

.github/workflows/update-release-branch.yml
```diff
@@ -144,6 +146,72 @@ jobs:
...
+                  print release_major_minor
+                  main_major_minor_count++
+                  next
+                }
+                /^export const currentAspireVersion = / {
+                  print release_version
+                  main_version_count++
+                  next
+                }
+                /^export const currentAspirePreviewVersion = / {
```
eerhardt, line 186: Does this need to be updated every time we add a line to the versions.mjs file?

.github/workflows/update-release-branch.yml
IEvangelist: Every time we ship a new rev of Aspire, patches included.

PR #1681, opened 2026-09-16
fix: preserve client navigation lifecycles across pages and deployments

IEvangelist:
## Summary

Follow-up to microsoft/aspire.dev#1675. Keep `<ClientRouter fallback="swap" />` and uninterrupted native PiP during normal browsing, while explicitly managing page-owned JavaScript and deployment changes.

Fresh production search and Twoslash work. The reported failures were reproduced by carrying the old runtime into newer HTML: the old search class remained registered, shortcuts referenced detached dialogs, and the old Twoslash workaround removed newly attached hover listeners. Fetching new HTML does not reset the running JavaScript environment.

### Changes

- Use Astro's preparation loader and cancellation API to reload across deployment versions. Same-version browsing remains client-side; deployment reloads intentionally close PiP.
- Preserve the selected theme and Astro history metadata across navigation.
- Dispose pivot and install-dialog listeners/timers with their owning page. Keep floating pivot controls below the responsive table of contents.
- **Remove `astro-mermaid`, `starlight-scroll-to-top`, and both proposed patches.** Render diagrams through the existing `mermaid` dependency and a small remark transform. Use a native scroll-to-top component instead of patching plugin internals.
- Keep diagram loading lazy, serialize theme/render work, discard stale completions, and preserve visible errors. Keep scroll controls localized and usable with mouse, touch, Enter, and Space.
- Configure 1DS to exclude `unload` without disabling other lifecycle flushing or click capture. Remove routine analytics chatter, not error reporting.
- Add regression coverage and document lifecycle ownership using the official Astro APIs.

No new dependency patches or compatibility framework are introduced. The pre-existing Starlight and icon-cache patches are unchanged.

The Edge lazy-image notice is informational. The evaluated-script `reportAllChanges`/`startTime` exception remains unattributed; this PR does not claim to fix it.

Links to Astro's official documentation as the implementation reference. No commercial or sponsored links are added.

## Validation

- 55 focused unit tests passed, including Markdown/MDX output and localized component rendering.
- 12 UI browser scenarios passed on Astro dev across desktop, tablet, and mobile, covering diagram themes, repeated navigation/history, and native scroll activation.
- Scoped TypeScript, typed ESLint, and frozen-lockfile installation passed. Remaining dependency versions and resolution metadata are unchanged.
- Full production rebuild and production-browser confirmation of the patch-free revision are in progress. Earlier production results from the patched revision are not claimed as validation of this replacement.


src/frontend/src/components/pivot-selector.ts
```diff
@@ -0,0 +1,282 @@
...
+
+export class PivotSelector extends HTMLElement {
+  private controller?: AbortController;
+
+  connectedCallback() {
+    if (this.controller) return;
+    this.controller = new AbortController();
+    const { signal } = this.controller;
+    // A swapped element connects before Astro updates the URL and restores scroll.
+    document.addEventListener('astro:page-load', () => this.initialize(signal), {
```
JamesNK: On a fresh document this now waits exclusively for `astro:page-load`, which Astro registers on `window.load`; the previous implementation initialized at `DOMContentLoaded`. A slow async resource can therefore leave the pivot buttons inert and every alternative visible until the resource finishes. Please initialize fresh documents at DOM readiness while retaining `astro:page-load` for swaps, and cover the case where an async resource remains pending past `DOMContentLoaded`.

src/frontend/src/scripts/mermaid.ts
```diff
@@ -0,0 +1,136 @@
...
+    (diagram) => renderedThemes.get(diagram) !== selectedTheme
+  );
+  if (!diagrams.length) return;
+  for (const diagram of diagrams) diagram.dataset.diagram ??= diagram.textContent ?? '';
+
+  let mermaid: Mermaid;
+  try {
+    mermaid = await loadMermaid();
+    if (!current()) return;
+    mermaid.initialize({ startOnLoad: false, theme: selectedTheme });
```
JamesNK, line 62: The removed integration configured light-mode diagrams with Mermaid's `forest` theme, while this maps light mode to `default`. Is that palette change intentional? Otherwise, preserve `forest` here and add a visual assertion so future runtime replacements retain the diagram appearance.

JamesNK (approved):
Reviewed at `a1a767962af710f40d1b34d9702e02ee84efddc5`. The exact-head production build, unit suite, and desktop/tablet/mobile Playwright runs are green. Approving with two inline follow-up comments covering fresh-load initialization timing and the light-mode Mermaid palette.

src/frontend/src/scripts/mermaid.ts
IEvangelist: Restored the prior light-mode Mermaid `forest` theme in `e270e183`; dark mode remains `dark`. The runtime unit coverage now explicitly asserts the `forest` → `dark` theme sequence across theme changes and stale-render cancellation.

<sub>Auto-replied by the [GitHub Copilot app](https://gh.io/app)</sub>

src/frontend/src/components/pivot-selector.ts
IEvangelist: Fixed in `e270e183`. Fresh documents now initialize the pivot selector at `DOMContentLoaded`, while swapped documents still wait for `astro:page-load`; a one-shot guard prevents duplicate initialization. Added coverage where page load remains pending after DOM readiness.

<sub>Auto-replied by the [GitHub Copilot app](https://gh.io/app)</sub>

src/frontend/tests/unit/site-ui-runtime.vitest.test.ts
```diff
@@ -0,0 +1,558 @@
...
+
+const mermaidSource = readFileSync(
+  new URL('../../src/scripts/mermaid.ts', import.meta.url),
+  'utf8'
+);
+const scrollComponent = readFileSync(
+  new URL('../../src/components/ScrollToTop.astro', import.meta.url),
+  'utf8'
+);
+const scrollSource = scrollComponent.match(/<script>([\s\S]*?)<\/script>/)?.[1];
```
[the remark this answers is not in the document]
IEvangelist: Fixed in `48b1a0f5`. The component-script extraction regex is now case-insensitive, with an explicit uppercase `<SCRIPT>` regression test.

<sub>Auto-replied by the [GitHub Copilot app](https://gh.io/app)</sub>

src/frontend/src/scripts/mermaid.ts
IEvangelist: Correction after inspecting `astro-mermaid@2.1.0` itself: with `autoTheme: true`, the plugin mapped explicit light mode to `default` and dark mode to `dark`; the configured `forest` value was only its unrecognized/missing-theme fallback. Commit `5c364338` now matches that effective `default`/`dark` behavior exactly, with the unit assertions updated.

<sub>Auto-replied by the [GitHub Copilot app](https://gh.io/app)</sub>

